PR #2, opened 2014-01-17
Setup and update

teeniimoodi:
Introduce `geordi` script which brings the powerful `setup` and `update` commands, plus `devserver`.

Move cuc, rs, tests, and migrate-all scripts there.


codener:
Was merged some days ago.

PR #26, opened 2016-08-24
Pass port to mysqldump in dumple.

enezhadian:
`dumple` command ignores the port for mysql database.


kratob:
Thank you. Change is released in 1.4.2.

PR #29, opened 2017-01-18
Use port from yaml file in the dump loader.

rin:


kratob:
Thanks!
Released as version 1.6.2.

PR #40, opened 2018-01-12
Remove `geordi purge-kernels`. Use `apt autoremove` instead.

triskweline:
These commits remove the `purge-kernels` command. We can use `apt autoremove` instead, which has tests. It's good to have tests for a command that can brick your computer.

This also encourages people to install `geordi` as `root`, which is never a good thing.


codener:
Prior to Ubuntu 16.04, `apt-get autoremove` would often fail to remove old kernels, that's why this command exists. It's a good thing to remove this now :+1:

codener:
Released as 1.7.0.
(I didn't feel like removing a discouraged command should be a major release.)

Issue #46, opened 2018-03-02
Add feature to drop unused databases


fheinle:
Old databases on local workstations should be deleted when no longer in use. Whitelisted databases should be excluded from deletion.

Should support both Postgres and MariaDB/MySQL.

PR #47, opened 2018-03-02
Dbcleanup integration

### LINKED ISSUE #45: Add feature to delete database dump files
Old `.dump` files should deleted in the directories geordi creates them, i.e. the current working directory and `~/dumps`.

fheinle:
Add features to remove old databases; both dump files and databases in local installations of MariaDB/MySQL and Postgres

Fixes #45 and #46

### DIFF
--- a/lib/geordi/commands/db_clean.rb
+++ b/lib/geordi/commands/db_clean.rb
@@ -1,0 +1,48 @@
+desc 'drop-databases', 'Delete local non-whitelisted databases'
+long_desc <<-LONGDESC
+
+Drop non-whitelisted databases from local installations of MySQL/MariaDB and
+  PostgreSQL. Offers to edit the whitelist.
+LONGDESC
+
+option :postgres_only, :aliases => '-P', :type => :boolean,
+  :desc => 'Only clean Postgres', :default => false
+option :mysql_only, :aliases => '-M', :type => :boolean,
+  :desc => 'Only clean MySQL/MariaDB', :default => false
+option :postgres, :banner => 'STRING',
+  :desc => 'Use Postgres port or socket'
+option :mysql, :banner => 'STRING',
+  :desc => 'Use MySQL/MariaDB port or socket'
+
+def drop_databases
+  require 'geordi/db_cleaner'
+  fail '-P and -M are mutually exclusive' if options.postgres_only and options.mysql_only
+  mysql_flags = nil
+  postgres_flags = nil
+
+  unless options.mysql.nil?
+    begin
+      mysql_port = Integer(options.mysql)
+      mysql_flags = "--port=#{mysql_port} --protocol=TCP"
+    rescue AttributeError
+      unless File.exist? options.mysql
+        fail "Path #{options.mysql} is not a valid MySQL socket"
+      end
+      mysql_flags = "--socket=#{options.mysql}"
+    end
+  end
+
+  unless options.postgres.nil?
+    postgres_flags = "--port=#{options.postgres}"
+  end
+
+  extra_flags = {'mysql' => mysql_flags,
+                 'postgres' => postgres_flags
+  }
+  cleaner = DBCleaner.new(extra_flags)
+  cleaner.clean_mysql unless options.postgres_only
+  cleaner.clean_postgres unless options.mysql_only
+
+  success 'Done.'
+end
+
--- a/lib/geordi/commands/db_clean.rb
+++ b/lib/geordi/commands/db_clean.rb
@@ -1,17 +1,29 @@
 desc 'drop-databases', 'Delete local non-whitelisted databases'
 long_desc <<-LONGDESC
 
-Drop non-whitelisted databases from local installations of MySQL/MariaDB and
-  PostgreSQL. Offers to edit the whitelist.
+Delete local MySQL/MariaDB and Postgres databases that are not whitelisted.
+
+Example: `geordi drop_databases`
+
+Check both MySQL/MariaDB and Postgres on the machine running geordi for databases
+and offer to delete them. Excluded are databases that are whitelisted. This comes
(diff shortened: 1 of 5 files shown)


codener:
Dear @fheinle, thanks for putting effort into this.
I've installed your two commands and tried them locally. They seem to do their job well!

However, I have some "workflow" hints that might make using the commands easier for newbies like me:
- The "Cleaning default directories" announcement might better read "Looking for *.dump files", followed by a `#note` that lists the checked directories. E.g. `> in ~/dumps and .`, or the given directory
- The "Dropping ... databases" announcement might sound dangerous. "Looking for ... databases" will be softer.
- For me, the (initial) list of databases was really long (>100 entries). I did not even see the heading and explanation, only the end of the list and a questioning "Proceed?". For users like me, it would be nice to have the "These databases are not whitelisted and ..." hint above the prompt. 
- The warning about a missing $HOME/.my.conf left me wondering what I need to do. What about a warn("You should create a ~/.my.cnf file first, or you'll need to enter your MySQL root password for each deleted database (see https://makandracards.com/makandra/1337)")

I have some more notes on the whitelist:
- I think it would be helpful if the whitelist began with a quick explanation like "The databases listed in this file will be excepted from deletion".
- I did not totally understand what "always whitelisted:" meant. Something like "These databases will never be deleted:" might be better understandable. I.e. I didn't even know there was a database called "mysql" or "postgres".
- I couldn't see my databases any more when the whitelist editor opened. Could you initialize the whitelist with all databases commented out? I would only need to comment my needful databases back in, and could delete the other rows.
- `foo_testd?` is missing a backslash `foo_test\d?`; if you keep it in a double-quoted string, you need to escape the \ with a \, i.e. have a double \\.
- Please add _cucumber suffix to the _development and _test suffices, in old projects we do have this one, too.

Thanks for documenting the commands. However, the README is auto-generated from the commands' descriptions with `rake readme`. Please merge your README changes into the commands' LONGDESC, so they won't get lost when someone recreates the README, and run `rake readme` afterwards.

As a side note, please do not list databases with ">", as the `note` helper is intended for comments and information, not for output.

---

What do you think about adding a `geordi tidy` command, that invokes `drop-database`, `delete-dumps`, `clean` and `remove-executable-flags`?

lib/geordi/commands/db_clean.rb
```diff
@@ -0,0 +1,47 @@
...
+  PostgreSQL. Offers to edit the whitelist.
+LONGDESC
+
+option :postgres_only, :aliases => '-P', :type => :boolean,
+  :desc => 'Only clean Postgres', :default => false
+option :mysql_only, :aliases => '-M', :type => :boolean,
+  :desc => 'Only clean MySQL/MariaDB', :default => false
+option :postgres, :banner => 'STRING',
+  :desc => 'Use Postgres port or socket'
+option :mysql, :banner => 'STRING',
```
codener: Please replace `STRING` with `PORT_OR_SOCKET`. Geordi's command help will be more self-explaining this way.

lib/geordi/db_cleaner.rb
```diff
@@ -0,0 +1,183 @@
...
+          if prompt('Edit the whitelist? [y]es or [n]o') == 'y'
+            proceed = 'e'
+          else
+            return []
+          end
+        end
+        if proceed.empty?
+          note "The following #{dbtype} databases are not whitelisted and will be deleted:"
+          deletable_dbs.sort.each do |db|
+            note db
```
codener: Please simply `puts` the database, as `note` is intended for comments and information, not output.

lib/geordi/db_cleaner.rb
```diff
@@ -0,0 +1,183 @@
...
+          end
+          proceed = prompt('Proceed? [y]es, [n]o or [e]dit whitelist')
+        end
+        case proceed
+        when 'e'
+          proceed = ''  # reset user selection
+          edit_whitelist dbtype
+        when 'n'
+          announce 'Not deleting anything'
```
codener: Please prefer `success` here, to signal a positive result.

lib/geordi/db_cleaner.rb
```diff
@@ -0,0 +1,183 @@
...
+    end
+    private :filter_whitelisted
+  end
+end
+
+def choose_texteditor
+  %w[$VISUAL $EDITOR /usr/bin/editor vi].each do |texteditor|
+    return texteditor if cmd_exists? texteditor
+  end
+end
```
codener: These methods are globally defined. Please prefer to either include them into the database cleaner class, or move them to Geordi::Util.

triskweline:
> What do you think about adding a geordi tidy command, that invokes drop-database, delete-dumps, clean and remove-executable-flags?

Let's not add more commands with generic words that delete your data. I'm already not using `clean` because I don't know what it does.

fheinle:
Thanks for your feedback, I have made the requested changes to my branch. As per the discussion here I have not added a `tidy` command. 

Notable changes in handling include editing the whitelist of databases that should not be dropped. Upon editing the whitelist, all database names will appear in the editor window, whitelisted names prefixed with `keep`, non-whitelisted names prefixed with `drop`, similar to how `git rebase` works. Edits to the whitelist will be saved to a temp file and only whitelisted db names will be saved to the config file, since the databases that are prefixed with `drop` will actually be dropped in the next step and keeping their names around didn't make much sense to me.

codener:
Thanks for elaborating on this. Some notes:

When invoking `geordi drop-databases`, I'm immediately prompted my sudo password. I would like a hint as before, what Geordi is doing, so I will understand why it needs sudo, e.g. `note 'Trying to authenticate MySQL as system user'` (or whatever the reason is ;)

The my.cnf hint has bad indentation in the second line. As for me, both lines could be independent invocations of `warn`.

codener:
Released as 1.9.0. Thanks @fheinle!

PR #60, opened 2019-01-23
Remove db name prefix when reading whitelist

### LINKED ISSUE #59: Filter duplicate 'keep' after users manually edited whitelist
# The Problem

`geordi drop-databases` fails if users edited their `$HOME/.config/geordi/whitelists/*.txt` manually instead of through geordi but still used the same syntax like when editing through geordi

# Reason

The syntax of geordi's database whitelist file on disk differs from what the user is asked to edit when using the `drop-databases` command. The file on disk only has one database name per line, sans any `keep` or `drop` prefixes. To avoid having to solve conflicts between two lines (and derivative database names like `keepme` and `keepme_development`), no `drop deleteme` lines are added to the *white*list.  When using the `edit` option during `geordi drop-databases` however, the user is presented with a concatenated list of both whitelisted database names (prefixed with `keep`) and database names that are not whitelisted but present on local database servers (prefixed with `drop`). 

If users use the `keep`/`drop` syntax when editing the whitelist files manually, geordi adds another `keep` prefix for interactive editing the whitelist, which subsequently fails when trying to validate the new whitelist.

# Solution

To allow both the recommended way of iteratively editing the whitelist after geordi presents the list of databases it would drop *and* manual editing, copying, etc of whitelist files on disk, geordi should check the whitelist file if lines a) have more than one word b) the first word in that case is `keep` and just throw out the `keep` word. A warning message to the user should be output once, since geordi will save the new whitelist in the correct format.

fheinle-mak:
If users manually added the `keep` prefix to the whitelist file on disk,
remove this prefix when reading the file to avoid syntax errors.

Fixes #59

### DIFF
--- a/lib/geordi/db_cleaner.rb
+++ b/lib/geordi/db_cleaner.rb
@@ -47,8 +47,31 @@ HEREDOC
         db.sub!(@derivative_dbname, '')
         tmpfile_content.push(['drop', db])
       end
-      whitelisted_dbs.each do |db|
-        tmpfile_content.push(['keep', db])
+      warn_manual_whitelist = false
+      whitelisted_dbs.each do |db_name|
+        # Remove 'keep' word from whitelist entries. This is not normally required since geordi
+        # does not save 'keep' or 'drop' to the whitelist file on disk but rather saves a list
+        # of all whitelisted db names and just presents the keep/drop information while editing
+        # the whitelist to supply users a list of databases they can whitelist by changing the
+        # prefix to 'keep'. Everything prefixed 'drop' is not considered whitelisted and thus
+        # not written to the whitelist file on disk.
+        #
+        # However, if users manually edit their whitelist files they might use the keep/drop
+        # syntax they're familiar with.
+        if db_name.start_with? 'keep '
+          db_name.gsub!(/keep /, '')
+          db_name = db_name.split[1..-1].join(' ')
+          warn_manual_whitelist = true
+        end
+        tmpfile_content.push(['keep', db_name]) unless db_name.empty?
+      end
+      if warn_manual_whitelist
+        warn <<-ERROR_MSG.gsub(/^\s*/, '')
+        Your whitelist #{whitelist} seems to have been generated manually.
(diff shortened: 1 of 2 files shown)


lib/geordi/db_cleaner.rb
```diff
@@ -47,9 +47,25 @@ def edit_whitelist(dbtype)
...
+        # However, if users manually edit their whitelist files they might use the keep/drop
+        # syntax they're familiar with.
+        if db_name.split.length > 1 && db_name.split.first == 'keep'
+          db_name = db_name.split[1..-1].join(' ')
+          warn_manual_whitelist = true
+        end
+        tmpfile_content.push(['keep', db_name]) unless db_name.empty?
       end
+      warn "Your whitelist #{whitelist} seems to have been generated manually. In that case, make sure to use only one database name per line and omit the 'keep' prefix.\n\n"\
+      'Launching the editor.' if warn_manual_whitelist
```
denzelem: Hier die gleiche Anmerkung wie in https://github.com/makandra/geordi/pull/62#discussion_r258378138.

lib/geordi/db_cleaner.rb
```diff
@@ -47,9 +47,25 @@ def edit_whitelist(dbtype)
...
+        # does not save 'keep' or 'drop' to the whitelist file on disk but rather saves a list
+        # of all whitelisted db names and just presents the keep/drop information while editing
+        # the whitelist to supply users a list of databases they can whitelist by changing the
+        # prefix to 'keep'. Everything prefixed 'drop' is not considered whitelisted and thus
+        # not written to the whitelist file on disk.
+        #
+        # However, if users manually edit their whitelist files they might use the keep/drop
+        # syntax they're familiar with.
+        if db_name.split.length > 1 && db_name.split.first == 'keep'
+          db_name = db_name.split[1..-1].join(' ')
```
denzelem: Man könnte auch `db_name.start_with?('keep ')` und `db_name.gsub(/\keep /, '')` sagen.

denzelem:
This was fixed in 2eb45a11681159cfc0f55635e1bb9c66779ae9e4.

PR #74, opened 2019-11-14
Show adapter type before asking for the DB password

dastra-mak:
When you are working on different projects that are using different kinds of databases, I think it would be helpful for the developer to be presented the kind of database he is asked to provide the password for.

### DIFF
--- a/lib/geordi/commands/create_database_yml.rb
+++ b/lib/geordi/commands/create_database_yml.rb
@@ -6,10 +6,12 @@ def create_database_yml
   if File.exists?(sample_yml) and not File.exists?(real_yml)
     announce 'Creating ' + real_yml
 
-    print 'Please enter your DB password: '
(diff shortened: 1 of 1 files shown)


lib/geordi/commands/create_database_yml.rb
```diff
@@ -6,10 +6,12 @@ def create_database_yml
   if File.exists?(sample_yml) and not File.exists?(real_yml)
     announce 'Creating ' + real_yml
 
-    print 'Please enter your DB password: '
+    sample = File.read(sample_yml)
+    adapter = sample.match(/adapter: (\w+)\n/).captures.first
+
+    print "Please enter your DB password for #{adapter}: "
```
codener: Please use this output:
```suggestion
    print "Please enter your #{adapter} password: "
```

Issue #52, opened 2018-11-05
Can we compress dumps before downloading?


triskweline:
I am working on several projects that have MySQL dumps with hundreds of megabytes of data. I need to routinely copy them to my local machine for debugging. Doing this with `geordi dump` takes several minutes.

I wondered if we could compress dumps before downloading? We can assume that `tar` and `gzip` is available on both remote server and local machine.

I think we once had a failed PR about that somewhere, but I can't find it anymore.

kratob:
Maybe using `scp -C` is enough? That should enable gzip compression in SSH.

PR #85, opened 2020-02-26
Compress remote dumps with gzip before downloading

triskweline:
Please review and release if OK.

### DIFF
--- a/lib/geordi/remote.rb
+++ b/lib/geordi/remote.rb
@@ -9,6 +9,7 @@ module Geordi
     include Geordi::Interaction
 
     REMOTE_DUMP_PATH = '~/dumps/dump_for_download.dump'
+    REMOTE_DUMP_PATH_GZ = "#{REMOTE_DUMP_PATH}.gz"
 
     def initialize(stage)
       @stage = stage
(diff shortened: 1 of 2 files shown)


lib/geordi/remote.rb
```diff
@@ -34,18 +35,21 @@ def select_server
...
+      destination_path_gz = "#{destination_path}.gz"
+
       relative_destination = Pathname.new(destination_path).relative_path_from Pathname.new(@config.root)
+      relative_destination_gz = "#{relative_destination}.gz"
 
-      note "Downloading remote dump to #{relative_destination} ..."
+      note "Downloading remote dump to #{relative_destination_gz} ..."
       server = @config.primary_server
-      Util.system! "scp -C #{ @config.user(server) }@#{ server }:#{REMOTE_DUMP_PATH} #{destination_path}"
+      Util.system! "scp -C #{ @config.user(server) }@#{ server }:#{REMOTE_DUMP_PATH_GZ} #{destination_path_gz} && gunzip -f #{destination_path_gz}"
```
denzelem: We already used the `-c` option within scp #52:

From the `scp` manual:

```
-C      Compression enable.  Passes the -C flag to ssh(1) to enable compression.
```

From the `ssh` manual:

```
-C      Requests compression of all data (including stdin, stdout, stderr, and data for
forwarded X11, TCP and UNIX-domain connections).  The compression algorithm is
the same used by gzip(1).  Compression is desirable on modem lines and other slow
connections, but will only slow down things on fast networks.  The default value can be
set on a host-by-host basis in the configuration files; see the Compression option.
```

@triskweline Do you think the manual compression before does provide any advantage?

triskweline:
I didn't realize we were already compressing, no need to gzip twice.

I added another commit to this branch that helped me in development. Can you take a look if that's worth merging?
https://github.com/makandra/geordi/pull/85/commits/e9aba4901924c7fcdf8d69bbb6f8d8ba6496c11e

This allowed me to test this change from my local geordi repository folder:

    ~/Projects/geordi > bundle exec geordi dump stagin RAILS_ENV=~/Projects/other-project

denzelem:
e9aba49 looks good to me. Thanks!

triskweline:
I only merged that one commit then. Thanks for the review!

PR #86, opened 2020-04-01
Fix false "geordi commit" message (closes #83)

### LINKED ISSUE #83: geordi commit falsely claims "no staged changes"
I have staged changes, but `geordi commit` warns "No staged changes. Will create an empty commit."

The issue is that Geordi only looks for added files:

https://github.com/makandra/geordi/blob/761782c9a4d44ba5ed89b8d88da071a146608d63/lib/geordi/util.rb#L97

It should instead check if any line starts with a letter. See documentation of [git status porcelain](https://git-scm.com/docs/git-status#_porcelain_format_version_2) and the used [short format](https://git-scm.com/docs/git-status#_short_format).

JohannaHartmann:
"geordi commit" would print a warning "no staged changes" unless new files were staged.
Now it respects **any** staged changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,6 +385,9 @@ calling your local geordi like so (adjust paths to your needs):
 
     # @option -I: add directory to load path
     ruby -I ../geordi/lib ../geordi/bin/geordi <command>
+    
+    # with debugger
+    ruby -r byebug -I ../geordi/lib ../geordi/bin/geordi <command>
 
 Don't forget to update this README. The whole `geordi` section is auto-generated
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -385,6 +385,9 @@ calling your local geordi like so (adjust paths to your needs):
 
     # @option -I: add directory to load path
     ruby -I ../geordi/lib ../geordi/bin/geordi <command>
+    
+    # with debugger
+    ruby -r byebug -I ../geordi/lib ../geordi/bin/geordi <command>
```
codener, line 390: Thanks for documenting this.

PR #100, opened 2020-06-24
Add helper class for settings and allow disabling vnc (closes #91)

### LINKED ISSUE #91: Make geordi aware of a headless Chrome driver
For current projects we don't use VNC anymore and use a headless Chrome instead. This makes it easier to interact with the Chrome window when debugging and removes one layer from the stack. 

It would be nice (imho) if Geordi does not wrap a headless Chrome in a VNC window. Projects with a headless Chrome define a tag `@no-headless` to start the Chrome visible for the user.

Can we detect if Geordi should use VNC or not easily? @codener @kratob @triskweline

niklas-hasselmeyer:
Wrote a settings class which reads from `.geordi.yml` and `~/.config/geordi/global.yml`.  
Updated `geordi commit` to use the new settings, deprecated the old `.pt_project_id` and `~/.gitpt` files.
Added an option to `geordi cucumber` to skip the vnc session via an option in `.geordi.yml`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,5 @@
 .idea
 pkg/*
 .DS_Store
-.pt_project_id
 tmp
 .byebug_history
--- a/lib/geordi/gitpt.rb
+++ b/lib/geordi/gitpt.rb
@@ -3,12 +3,14 @@ class Gitpt
   require 'highline'
   require 'tracker_api'
 
-  SETTINGS_FILE_NAME = '.gitpt'.freeze
-  PROJECT_IDS_FILE_NAME = '.pt_project_id'.freeze
+  # This require-style is to prevent Ruby from loading files of a different
+  # version of Geordi.
+  require File.expand_path('settings', __dir__)
 
   def initialize
     self.highline = HighLine.new
-    self.client = build_client(read_settings)
+    self.settings = Geordi::Settings.new
+    self.client = build_client
   end
 
   def run(git_args)
@@ -24,55 +26,15 @@ No staged changes. Will create an empty commit.
 
   private
 
-  attr_accessor :highline, :client
+  attr_accessor :highline, :client, :settings
 
-  def read_settings
-    file_path = File.join(ENV['HOME'], SETTINGS_FILE_NAME)
-
-    unless File.exist?(file_path)
-      highline.say HighLine::RESET
-      highline.say "Welcome to #{bold 'gitpt'}.\n\n"
-
-      highline.say highlight('Your settings are missing or invalid.')
-      highline.say "Please configure your Pivotal Tracker access.\n\n"
-      token = highline.ask bold('Your API key:') + ' '
-      highline.say "\n"
-
-      settings = { token: token }
-      File.open(file_path, 'w') do |file|
-        file.write settings.to_yaml
-      end
-    end
-
-    YAML.load_file(file_path)
-  end
-
-  def build_client(settings)
-    TrackerApi::Client.new(token: settings.fetch(:token))
+  def build_client
+    TrackerApi::Client.new(token: settings.pivotal_tracker_api_key)
   end
 
   def load_projects
-    project_ids = read_project_ids
+    project_ids = settings.pivotal_tracker_project_ids
     project_ids.collect { |project_id| client.project(project_id) }
-  end
-
-  def read_project_ids
-    file_path = PROJECT_IDS_FILE_NAME
-
-    if File.exist?(file_path)
-      project_ids = File.read('.pt_project_id').split(/[\s]+/).map(&:to_i)
-    end
-
-    if project_ids && (project_ids.size > 0)
-      project_ids
-    else
-      Geordi::Interaction.warn "Sorry, I could not find a project ID in #{file_path} :("
-      puts
-
-      puts "Please put at least one Pivotal Tracker project id into #{file_path} in this directory."
-      puts 'You may add multiple IDs, separated using white space.'
-      exit 1
-    end
   end
 
   def applicable_stories
@@ -132,10 +94,6 @@ No staged changes. Will create an empty commit.
 
   def bold(string)
     HighLine::BOLD + string + HighLine::RESET
-  end
-
-  def highlight(string)
-    bold HighLine::BLUE + string
   end
 
 end
--- a/lib/geordi/settings.rb
+++ b/lib/geordi/settings.rb
@@ -1,0 +1,155 @@
+require 'yaml'
+
+require File.expand_path('util', __dir__)
+require File.expand_path('interaction', __dir__)
+
+module Geordi
+  class Settings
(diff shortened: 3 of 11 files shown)


lib/geordi/gitpt.rb
```diff
@@ -2,12 +2,14 @@ class Gitpt
   require 'yaml'
   require 'highline'
   require 'tracker_api'
+  require 'geordi/settings'
```
codener, line 5: Please require settings just like in cucumber.rb. While this ^ seems to work, it will require the settings of the Geordi in Ruby's load path, which is not always the same file as the settings file relative to this one.

.gitignore
```diff
@@ -4,5 +4,6 @@
 pkg/*
 .DS_Store
 .pt_project_id
+.geordi.yml
```
codener, line 7: If you need to ignore a `.geordi.yml` generated by tests, maybe the test is also writing to the homedir? That's not too good. Please set the setting file paths in geordi/settings.rb depending on `ENV['GEORDI_TESTING']` (set in features/support/env.rb), and have them point to ./tmp/{local_settings,global_settings}.yml in tests.

lib/geordi/settings.rb
```diff
@@ -0,0 +1,83 @@
+class GeordiSettings
+  require 'yaml'
+
+  GLOBAL_SETTINGS_FILE_NAME = '.config/geordi/global.yml'.freeze
+  LOCAL_SETTINGS_FILE_NAME = '.geordi.yml'.freeze
+
+  ALLOWED_GLOBAL_SETTINGS = %w[ pivotal_tracker_api_key ].freeze
+  ALLOWED_LOCAL_SETTINGS = %w[ headless_browser pivotal_tracker_project_id ].freeze
```
codener, line 8: Please pluralize `pivotal_tracker_project_ids`.

lib/geordi/gitpt.rb
```diff
@@ -24,27 +26,32 @@ def run(git_args)
...
+        token = highline.ask bold('Your API key:') + ' '
+        highline.say "\n"
       end
+
+      settings.pivotal_tracker_api_key = token
     end
 
-    YAML.load_file(file_path)
+    { token: settings.pivotal_tracker_api_key }
   end
```
codener, line 55: Please move this whole function to `Geordi::Settings`.

lib/geordi/gitpt.rb
```diff
@@ -57,19 +64,26 @@ def load_projects
...
-      Geordi::Interaction.warn "Sorry, I could not find a project ID in #{file_path} :("
+      Geordi::Interaction.warn "Sorry, I could not find a project ID in .geordi.yml :("
       puts
 
-      puts "Please put at least one Pivotal Tracker project id into #{file_path} in this directory."
+      puts "Please put at least one Pivotal Tracker project id into the .geordi.yml file in this directory, e.g:\n"
+      puts "pivotal_tracker_project_id: 123456\n"
       puts 'You may add multiple IDs, separated using white space.'
       exit 1
     end
```
codener, line 89: Please move this whole function to `Geordi::Settings`. It should always return an array of project ids, and may serialize it as Yaml array OR white-space separated string.

lib/geordi/gitpt.rb
```diff
@@ -28,67 +28,15 @@ def run(git_args)
...
-      settings.pivotal_tracker_api_key = token
-    end
-
-    { token: settings.pivotal_tracker_api_key }
-  end
-
-  def build_client(settings)
-    TrackerApi::Client.new(token: settings.fetch(:token))
+  def build_client
+    TrackerApi::Client.new(token: settings.pivotal_tracker_api_key)
```
codener, line 32: :ok_hand:

lib/geordi/settings.rb
```diff
@@ -1,83 +1,153 @@
...
+    require 'yaml'
 
-  GLOBAL_SETTINGS_FILE_NAME = '.config/geordi/global.yml'.freeze
-  LOCAL_SETTINGS_FILE_NAME = '.geordi.yml'.freeze
+    require File.expand_path('util', __dir__)
 
-  ALLOWED_GLOBAL_SETTINGS = %w[ pivotal_tracker_api_key ].freeze
-  ALLOWED_LOCAL_SETTINGS = %w[ headless_browser pivotal_tracker_project_id ].freeze
+    GLOBAL_SETTINGS_FILE_NAME = Util.testing? ? './tmp/global_settings.yml'.freeze : File.join(ENV['HOME'], '.config/geordi/global.yml').freeze
+    LOCAL_SETTINGS_FILE_NAME = Util.testing? ? './tmp/local_settings.yml'.freeze : './.geordi.yml'.freeze
```
codener, line 8: :ok_hand:

lib/geordi/settings.rb
```diff
@@ -1,83 +1,153 @@
...
 
-  def initialize
-    read_settings
-  end
+    GITPT_FILE_NAME = '.gitpt'.freeze
+    PROJECT_IDS_FILE_NAME = '.pt_project_id'.freeze
```
codener, line 11: Please inline these deprecated filenames (i.e. don't store them as constants). They will not be changed, but their functionality removed altogether at some point.

lib/geordi/settings.rb
```diff
@@ -1,83 +1,153 @@
...
-      global_settings = YAML.safe_load(File.read(global_path))
-      check_for_invalid_keys(global_settings, ALLOWED_GLOBAL_SETTINGS, global_path)
+      case project_ids
+      when Array
+        return project_ids if project_ids.size > 0
+      when String
+        return project_ids.split(/[\s]+/).map(&:to_i)
+      when Integer
+        return [project_ids]
+      end
```
codener, line 47: The first `when` will return `nil` on empty array. This will crash `Geordi::Gitpt#load_projects` https://github.com/makandra/geordi/pull/100/files#diff-8e55a5101bab5a11fa81cb519817f640R37

Please do the normalization to array as you're doing now (except for that `if` in L42), but avoid the early returns. Instead, add an `if` block below the `case` that
- returns available project ids or
- exits with the warning below when empty

codener:
Ok, so I installed Geordi 3.2.0 and got a bunch of errors on `geordi commit`. Please install yourself (checkout your branch, switch to the Ruby version of the project you want to try Geordi in and `rake install`) and resolve them. Hints: 

- Missing "highline" in settings.rb (please extract all current requires to the top of the file. That's ok since settings are only selectively required, and always need those libraries.
- Missing methods (e.g. `highlight`) in settings.rb. If possible, please adhere to the current interaction pattern of `announce` + (maybe some `note`s) + command output. You find Geordi's interaction library in `Geordi::Interaction`, which you may simply `include` in another class (see usage throughout Geordi). You are encouraged to prefer existing interaction methods from that module over custom output. If you want to see examples of these interactions, try `geordi deploy` or `geordi cucumber -v`.

Please also validate other affected commands perform as expected (i.e. `cucumber`), then push an update.

PR #102, opened 2020-06-30
Let --select-server option take the number of the server to skip the …

### LINKED ISSUE #43: Let --select-server option take the number of the server to skip the menu
The `shell` and `console` commands have a `--select-server` option that will bring up a menu asking which of the servers to connect to.

Desired behavior:

```
$ geordi shell --select-server 3
> Geordi opens a shell on the 3rd server
$ geordi shell --select-server
> When called without a number, the server menu should open as before.
```

This makes using these commands more flexible.

JohannaHartmann:
…menu (closes #43)

The server number can now be passed to the `--select-server` option for the `geordi shell` and `geordi console` commands to connect directly to the server, e.g. `geordi shell staging --select-server 2`.
Thereby the server select menu will be skipped.
The alias `-s` can still be used.
The server select menu will open as usual with a command like `geordi shell staging --select-server`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+- Improvement #43: `--select-server` option on `geordi shell` and `geordi console` can take the number of the server to connect to it directly and to skip the menu.
+
 ### Compatible changes
 
 ### Breaking changes
--- a/features/shell.feature
+++ b/features/shell.feature
@@ -4,7 +4,7 @@ Feature: The shell command
     Given a file named "Capfile" with "Capfile exists"
 
 
-  Scenario: It opens a remote shell on the primary server
+  Scenario: It opens a remote shell on the primary server when no option is given
     Given a file named "config/deploy.rb" with "deploy.rb exists"
     And a file named "config/deploy/geordi.rb" with:
     """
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+- Fix #43: `--select-server` option on `geordi shell` and `geordi console` can take the number of the server to connect to it directly and to skip the menu.
```
codener, line 9: I'd not call this a fix. What about "Improvement"?

features/shell.feature
```diff
@@ -17,6 +17,56 @@ Feature: The shell command
...
+    set :deploy_to, '/var/www/example.com'
+    server 'first.example.com'
+    server 'second.example.com'
+    """
+
+    When I run `geordi shell geordi --select-server` interactively
+    # Answer prompt "Connect to? [1]"
+      And I type "2"
+    Then the output should contain "# Opening a shell on geordi"
+    Then the output should contain "Util.system! ssh, deploy@second.example.com, -t, cd /var/www/example.com/current && bash --login"
```
codener, line 34: :ok_hand:

features/shell.feature
```diff
@@ -17,6 +17,56 @@ Feature: The shell command
...
+    set :user, 'deploy'
+    set :deploy_to, '/var/www/example.com'
+    server 'first.example.com'
+    server 'second.example.com'
+    """
+
+    When I run `geordi shell geordi --select-server 1`
+    Then the output should contain "Util.system! ssh, deploy@first.example.com, -t, cd /var/www/example.com/current && bash --login"
+
+    When I run `geordi shell geordi -s 2`
```
codener, line 50: Would it work as `geordi shell geordi -s2`? If so, please use that in the test.

lib/geordi/commands/shell.rb
```diff
@@ -2,12 +2,16 @@
 long_desc <<-LONGDESC
 Example: `geordi shell production`
 
-Lets you select the server to connect to when called with `--select-server`:
+Lets you select the server to connect to from a menu when called with `--select-server` or the alias `-s`:
 
     geordi shell production -s
+
+Lets you select the server to connect to directly by passing the number:
```
codener, line 9: In a way, this is "selecting" as well, but I'm not sure everyone will understand. What about: "If you already know the number of the server you want to connect to, just pass it along:"

Again, if `-s2` works, please use the syntax here.

lib/geordi/commands/dump.rb
```diff
@@ -21,7 +21,7 @@
 DESC
 
 option :load, aliases: ['-l'], type: :string, desc: 'Load a dump'
-option :select_server, default: false, type: :boolean, aliases: '-s'
+option :select_server, type: :string, aliases: '-s'
```
codener, line 24: Why is there a `select_server` option to `dump`, but lib/geordi/remote.rb:37 states that "dump must be generated on the primary server"? Please check this. If the `select_server` option to `dump` is obsolete, please remove it and also remove remote.rb:37 in a separate Commit.

features/shell.feature
```diff
@@ -17,6 +17,56 @@ Feature: The shell command
...
+    Then the output should contain "Util.system! ssh, deploy@second.example.com, -t, cd /var/www/example.com/current && bash --login"
+
+
+  Scenario: It opens a remote shell on the primary server when the server number is invalid
```
codener, line 54: On invalid server number, please do not use the primary server but print a warning (`Interaction.warn`) and open the menu instead. Else the user easily accesses an unexpected server!

lib/geordi/commands/shell.rb
codener: Please remember to update the README when changing this.

codener: Please also add this description to console.rb.

PR #104, opened 2020-07-08
Fix tests

niklas-hasselmeyer:
Since PR #100 had been merged, tests were red. This commit fixes this and adds a test for the new feature.

### DIFF
--- a/features/commit.feature
+++ b/features/commit.feature
@@ -1,5 +1,7 @@
 @ruby>=2.1
 Feature: Creating a git commit from a Pivotal Tracker story
+  Background:
+    Given a file named "tmp/global_settings.yml" with "pivotal_tracker_api_key: my_api_key"
 
   Scenario: Extra arguments are forwarded to "git commit"
     Given I have staged changes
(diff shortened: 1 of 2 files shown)


features/commit.feature
```diff
@@ -15,3 +17,10 @@ Feature: Creating a git commit from a Pivotal Tracker story
       # No optional message
       And I type ""
     Then the output should contain "> No staged changes. Will create an empty commit."
+
+  Scenario: With an empty global config file, the user is prompted for their PT API key
+    Given a file named "tmp/global_settings.yml" with "--- {}\n"
+    When I run `geordi commit` interactively
+      And I type "abc"
+      And I type ""
+    Then the file "tmp/global_settings.yml" should contain "pivotal_tracker_api_key: abc"
```
codener, line 26: Please do not clear the settings file, but remove it instead. This test covers the current situation of all Geordi users lacking the global settings file. The scenario should be titled "Without a global config file ...".

A test should explain itself as best as possible. Please replace "abc" with "my tracker api key" and fill in the empty quotes with "optional message".

lib/geordi/settings.rb
```diff
@@ -114,7 +114,7 @@ def save_global_settings
     # deprecated
     def gitpt_api_key_old
       file_path = File.join(ENV['HOME'], '.gitpt')
-      if File.exist?(file_path)
+      if File.exist?(file_path) && !Util.testing?
```
codener, line 117: Ok. It is hacky to disable the method this way, but there is no way around – the test cannot rely on files in the developer's `~`, and after all, it is deprecated.

PR #105, opened 2020-07-22
Fix geordi console and geordi deploy when not bundled

niklas-hasselmeyer:


lib/geordi/commands/deploy.rb
```diff
@@ -94,6 +94,8 @@ def deploy(target_stage = nil)
     git_call << "git merge #{source_branch}" if merge_needed
     git_call << 'git push' if push_needed
 
+    invoke_cmd 'bundle_install'
+
```
codener, line 98: This change should not be necessary, as there is no Bundler invocation between L98 and L107. The Capistrano call is executed in L108

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Compatible changes
 - Improved documentation; README now includes command options.
+- Fix #90: `geordi console` and `geordi deploy` now work correctly if the project hasn't been bundled before
```
codener, line 12: Please update this as required by changes below.

lib/geordi/commands/deploy.rb
niklas-hasselmeyer: There is a call to `Util.gem_major_version` in L98/L100 which uses `bundle list`

codener: :detective: ok!

PR #107, opened 2020-07-22
Respect binstubs (closes #94)

### LINKED ISSUE #94: Geordi should use binstubs if present
Binstubs that respect Spring can e.g. start tests faster, so `geordi cucumber` and `bundle rspec` should use the respective binstub if present.

Please check if more binstubs should be used.

niklas-hasselmeyer:
Now geordi uses binstubs for rails, rake, and cucumber, if present. Binstubs for rspec were already implemented.

### DIFF
--- a/features/migrate.feature
+++ b/features/migrate.feature
@@ -4,8 +4,8 @@ Feature: The migrate command
     Given a directory named "db/migrate"
 
   Scenario: Migrating with parallel_tests uses the rake binstub if it exists
-    Given an empty file named "bin/rake"
+    Given a file named "bin/rake" with "binstub"
       And a file named "Gemfile" with "gem parallel_tests"
 
-    When I run `geordi migrate` interactively
+    When I run `geordi migrate`
     Then the output should contain "Util.system! bin/rake, db:migrate parallel:prepare"
--- a/features/migrate.feature
+++ b/features/migrate.feature
@@ -1,0 +1,11 @@
+Feature: The migrate command
+
+  Background:
+    Given a directory named "db/migrate"
+
+  Scenario: Migrating with parallel_tests uses the rake binstub if it exists
+    Given an empty file named "bin/rake"
+      And a file named "Gemfile" with "gem parallel_tests"
+
+    When I run `geordi migrate` interactively
+    Then the output should contain "Util.system! bin/rake, db:migrate parallel:prepare"
--- a/features/rake.feature
+++ b/features/rake.feature
@@ -1,7 +1,10 @@
 Feature: The rake command
 
-  Scenario: When a bin/rake file exists, it is executed
-    Given an empty file named "bin/rake"
+  Background:
+    Given a file named "config/environments/development.rb" with "enable environment"
 
-    When I run `geordi rake` interactively
+  Scenario: A rake binstub is used if present
+    Given a file named "bin/rake" with "binstub"
+
+    When I run `geordi rake`
     Then the output should contain "Util.system! bin/rake"
--- a/features/rake.feature
+++ b/features/rake.feature
@@ -1,0 +1,7 @@
+Feature: The rake command
+
+  Scenario: When a bin/rake file exists, it is executed
+    Given an empty file named "bin/rake"
+
+    When I run `geordi rake` interactively
+    Then the output should contain "Util.system! bin/rake"
--- a/features/rspec.feature
+++ b/features/rspec.feature
@@ -1,17 +1,17 @@
 Feature: The rspec command
 
   Background:
-    Given an empty file named "spec/spec_helper.rb"
+    Given a file named "spec/spec_helper.rb" with "enable rspec"
 
-  Scenario: When a bin/rspec file exists, it is executed
-    Given an empty file named "bin/rspec"
+  Scenario: A rspec binstub is used if present
+    Given a file named "bin/rspec" with "binstub"
 
-    When I run `geordi rspec` interactively
-    Then the output should contain "Util.system! bundle exec bin/rspec"
+    When I run `geordi rspec`
+    Then the output should contain "Util.system! bin/rspec"
 
-  Scenario: When a bin/rake file exists, it is used to run parallel tests
-    Given an empty file named "bin/rake"
-      And a file named "Gemfile" with "gem parallel_tests"
+  Scenario: A rake binstub is used to run parallel tests if present
+    Given a file named "bin/rake" with "binstub"
+      And a file named "Gemfile" with "gem 'parallel_tests'"
 
-    When I run `geordi rspec` interactively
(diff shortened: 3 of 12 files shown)


lib/geordi/util.rb
```diff
@@ -62,21 +62,29 @@ def system!(*commands)
         end
       end
 
+      def binstub(executable, arguments='')
+        binstub_file = "bin/#{executable}"
+
+        command = 'bundle exec '
+        command << File.exists?(binstub_file) ? binstub_file : executable
+        command << ' ' + arguments
```
codener, line 70: Elegant solution :+1: 

I forgot to mention: `bundle exec` is not required when using binstubs as well.

lib/geordi/util.rb
```diff
@@ -62,21 +62,29 @@ def system!(*commands)
...
+      end
+
       def console_command(environment)
         if gem_major_version('rails') == 2
           'script/console ' + environment
         elsif gem_major_version('rails') == 3
-          'bundle exec rails console ' + environment
+          binstub 'rails', "console #{environment}"
```
codener, line 77: If you define the binstub method as `def binstub(executable, *args)`, it can be used flexibly here like `binstub 'rails', 'console', environment`. wdyt?

lib/geordi/commands/unit.rb
```diff
@@ -5,7 +5,7 @@ def unit
     invoke_cmd 'yarn_install'
 
     Interaction.announce 'Running Test::Unit'
-    Util.system! 'bundle exec rake test'
+    Util.system! Util.binstub('rake', 'test')
```
codener, line 8: I would prefer invocation as `Util.system! Util.binstub('rake'), 'test'`, because only `rake` is the binstub here and `test` can be handled by `system!`.

lib/geordi/commands/rspec.rb
```diff
@@ -14,7 +14,7 @@ def rspec(*files)
 
     if Util.file_containing?('Gemfile', /parallel_tests/) && files.empty?
       Interaction.note 'All specs at once (using parallel_tests)'
-      Util.system! 'bundle exec rake parallel:spec', fail_message: 'Specs failed.'
+      Util.system! Util.binstub('rake', 'parallel:spec'), fail_message: 'Specs failed.'
```
codener, line 17: Same here.

lib/geordi/commands/rake.rb
```diff
@@ -12,7 +12,7 @@
 def rake(*args)
   %w[development test cucumber].each do |env| # update long_desc when changing this
     if File.exist? "config/environments/#{env}.rb"
-      call = %w[bundle exec rake] + args + ["RAILS_ENV=#{env}"]
+      call = Util.binstub('rake') + args + ["RAILS_ENV=#{env}"]
```
codener, line 15: I doubt this works as expected, since you're adding String + Array + Array. You should wrap the first summand with brackets `[]`.

lib/geordi/commands/migrate.rb
```diff
@@ -17,7 +17,7 @@ def migrate
       Interaction.note 'Development and parallel test databases'
       puts
 
-      Util.system! 'bundle exec rake db:migrate parallel:prepare'
+      Util.system! Util.binstub('rake', 'db:migrate parallel:prepare')
```
codener, line 20: Same here.

features/migrate.feature
```diff
@@ -0,0 +1,11 @@
+Feature: The migrate command
+
+  Background:
+    Given a directory named "db/migrate"
+
+  Scenario: Migrating with parallel_tests uses the rake binstub if it exists
+    Given an empty file named "bin/rake"
+      And a file named "Gemfile" with "gem parallel_tests"
+
+    When I run `geordi migrate` interactively
```
codener, line 10: I would not expect this to run interactively. Could you remove it?

features/migrate.feature
```diff
@@ -0,0 +1,11 @@
+Feature: The migrate command
+
+  Background:
+    Given a directory named "db/migrate"
+
+  Scenario: Migrating with parallel_tests uses the rake binstub if it exists
```
codener, line 6: What about migrations without parallel_test? Don't they have Rake involved?

features/rake.feature
```diff
@@ -0,0 +1,7 @@
+Feature: The rake command
+
+  Scenario: When a bin/rake file exists, it is executed
```
codener, line 3:
```suggestion
  Scenario: A rake binstub is used if present
```

features/rake.feature
```diff
@@ -0,0 +1,7 @@
...
+    Given an empty file named "bin/rake"
+
+    When I run `geordi rake` interactively
```
codener, line 6: Same here, should not be interactively.

features/rspec.feature
```diff
@@ -0,0 +1,17 @@
+Feature: The rspec command
+
+  Background:
+    Given an empty file named "spec/spec_helper.rb"
```
codener, line 4: Please do not explicitly create empty files here, but create them with a short hint of their intention instead.

```suggestion
    Given a file named "spec/spec_helper.rb" with "enable rspec"
```
(or whatever the reason is why you need this file)

Cucumber is an abstract language and should describe expected behavior (there is a file). It should not be used to "program" (like creating an empty file because its content does not matter).

features/rake.feature
codener: Please fix this everywhere.

features/rake.feature
codener: Please rename other scenarios likewise

features/rspec.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  Scenario: When a bin/rspec file exists, it is executed
+    Given an empty file named "bin/rspec"
+
+    When I run `geordi rspec` interactively
+    Then the output should contain "Util.system! bundle exec bin/rspec"
```
codener, line 10: A binstub should not be invoked with `bundle exec`.

lib/geordi/util.rb
```diff
@@ -62,21 +62,28 @@ def system!(*commands)
         end
       end
 
+      def binstub(executable, *arguments)
+        binstub_file = "bin/#{executable}"
+
+        command = File.exists?(binstub_file) ? [binstub_file] : ['bundle exec', executable]
+        command.append(*arguments).join(' ')
```
codener, line 69: Again, I believe `append` does not work in older Rubies. Please use `push` instead:

```suggestion
        command.push(*arguments)
        command.join(' ')
```

codener:
All tests are green, right?

Please install Geordi locally by switching the Ruby version and running `rake install`. Then, try the commands you modified, to see if they behave as expected.

PR #116, opened 2020-09-11
added chromedriver update to cucumber command

brunosedler:
The duration of the update_chromedriver command (if installation is skipped) is between 133 and 140ms (measured with `time`  bash command). Without checking for version equality it was about 116 to 118 ms, so the equality check takes about 20ms

### DIFF
--- a/lib/geordi/commands/chromedriver_update.rb
+++ b/lib/geordi/commands/chromedriver_update.rb
@@ -4,6 +4,10 @@ Example: `geordi chromedriver_update`
 
 This command will find and install the matching chromedriver for the currently
 installed Chrome.
+
+Setting `auto_update_chromedriver` to `true` in your global Geordi config file 
+(`~/.config/geordi/global.yml`), will automatically update chromedriver before 
+cucumber tests, in case Chrome and chromedriver versions don't match
 LONGDESC
 
 option :quiet_if_matching, type: :boolean, default: false,
--- a/lib/geordi/commands/cucumber.rb
+++ b/lib/geordi/commands/cucumber.rb
@@ -55,7 +55,9 @@ def cucumber(*args)
 
     invoke_geordi 'bundle_install'
     invoke_geordi 'yarn_install'
-    invoke_geordi 'chromedriver_update', quiet_if_matching: true
+    if settings.auto_update_chromedriver
+      invoke_geordi 'chromedriver_update', quiet_if_matching: true
+    end
 
     cmd_opts, files = args.partition { |f| f.start_with? '-' }
     cmd_opts << '--format' << 'pretty' << '--backtrace' if options.debug
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   ```
 
 ### Compatible changes
+
+* added `auto_update_chromedriver` as global setting option to automatically update chromedriver before cucumber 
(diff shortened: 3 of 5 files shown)


lib/geordi/commands/cucumber.rb
```diff
@@ -55,6 +55,7 @@ def cucumber(*args)
 
     invoke_geordi 'bundle_install'
     invoke_geordi 'yarn_install'
+    invoke_geordi 'chromedriver_update', show_warning: false
```
codener, line 58: Ok, we won't add a test because it is just too hard to fake the Chrome/chromedriver version determination.

lib/geordi/chromedriver_updater.rb
```diff
@@ -5,12 +5,12 @@
...
   class ChromedriverUpdater
 
-    def run
+    def run(options)
       chrome_version = determine_chrome_version
       chromedriver_version = determine_chromedriver_version
 
       if skip_update?(chrome_version, chromedriver_version)
-        Interaction.warn("No update required, you are using for both executables the same version #{chrome_version}!")
+        Interaction.warn("No update required, you are using for both executables the same version #{chrome_version}!") if options.show_warning
```
codener, line 13: Please turn this into an `Interaction.note`, as it is nothing to worry about.

lib/geordi/commands/chromedriver_update.rb
```diff
@@ -6,8 +6,11 @@
 installed Chrome.
 LONGDESC
 
+option :show_warning, type: :boolean, default: true,
+  desc: 'Show a warning if the chrome and chromedriver versions are already the same and nothing needs to be done.'
```
codener, line 10: I have objections regarding the option naming. "show warning" could be about anything. Let's get a bit closer to the actual thing we want here. What about `quiet_if_matching`? Default would be false, and the "no update" note would not be printed if the option is true.

Please also shorten the option description to about half its current length. This keeps it from wrapping when printed on the bash (confer `geordi help cucumber` > debug option description).

codener (changes requested):

codener (changes requested):
Thanks for the polishing! Please drop the question mark in the new option, then merge to master.

We'll not release it instantly, but instead use it locally for a while to see how it behaves.

codener (approved):
Fine, merge it!

codener:
@brunosedler I just found an important case we missed: Some users of Geordi may not be using chromedriver at all. Running `chromedriver-update` would break the Cucumber command for them. I see two possible solutions:

* Add a config option `autoupdate-chromedriver` for `.geordi.yml` and document it. The auto-updating should only take place when the option is set.
* Check whether `chromedriver` is present in the system, and run the update check if it is.

What do you think?

brunosedler:
@codener I think I like the first option better. The second one would add some more time to running the features, although we might have to check how long it takes. But if you don't have chromedriver it is annoying that it takes longer due to something you don't even need. So I like the "opt in" option better. Especially because we already thought about showing a message and ask permission to run the update, so with option one it would be something a user agrees to.

codener:
Checking for a `chromedriver` executable takes around 3ms – that's almost instant. Still a secret update might not be desirable for all users. Lets go with the config option.

- The option should be called "autoupdate-chromedriver". Please choose the casing (spinal/camel/snake) so that it does not need to be quoted in the yml file.
- Please document the option in the `long_desc` of the `chromedriver-update` command.
- Rebase on `master`
- Please add an entry to the CHANGELOG describing the new feature

codener (approved):
Super Sach!

PR #117, opened 2020-09-17
Fix filename corruption bug of geordi cucumber --modified (closes #115)

### LINKED ISSUE #115: geordi cucumber --modified corrupts file names
When running `geordi cucumber --modified`, Cucumber fails with "No such file or directory tures/pages.feature". Obviously, there is "fea" missing.

JohannaHartmann:
Fixed that `geordi cucumber --modified` corrupts the filename and cuts off `fea` of `feature`, which results in errors like:
```
No such file or directory tures/pages.feature
```

Also required the `English` gem, for correct (future) usage of global variables, like `$INPUT_RECORD_SEPARATOR`.
For reasons of consistency, however, this variable was replaced with `"\n"`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
+- Fixed #115: `geordi cucumber --modified` command, that corrupts filenames like:
+  ```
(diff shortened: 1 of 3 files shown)


codener (changes requested):
Hi @JohannaHartmann, thanks for fixing this! I have two remarks:

* Please do not add the `English` Gem as you don't use it. When developing Gems, we should be especially wary of adding new dependencies.
* Have you checked the whole project for usages of `$INPUT_RECORD_SEPARATOR` and the like?

Issue #124, opened 2020-10-26
"geordi vnc" is incompatible with tightvnc [20.04]


makmic:
Geordi assumes that `vnc4server` is used as the vncserver and uses its configuration flags for commands like `geordi cucumber`.
`vncserver` is no longer part of the 20.04 PPA and can be replaced with `tightvncserver`. They are both wrappers for `Xvnc`, but while 18.04 is shipped with "Xvnc Free Edition 4.x", 20.04 comes with "Xvnc version TightVNC-1.x".

## issues
1) The option `SecurityTypes` does not exist in the TightVNC implementation, but is used [by geordi](https://github.com/makandra/geordi/blob/master/lib/geordi/cucumber.rb#L14)
2) The [setup instructions](https://github.com/makandra/geordi/blob/master/lib/geordi/commands/_setup_vnc.rb#L26) from `geordi vnc` are not correct in 20.04

## References
* Documentation of the "old" Xvnc: https://linux.die.net/man/1/xvnc
* Documentation for the "new" Xvnc: http://manpages.ubuntu.com/manpages/trusty/man1/Xtightvnc.1.html

I'll try to work around these issues and provide a PR.

PR #125, opened 2020-10-26
[#124] Add tightvnc compatibility

makmic:
This commit switches from a "no auth" attempt to using a hardcoded
VNC password. This is neccessary to support both vnc4server and
tightvncserver.

The file `lib/geordi/vnc_password.txt` was created using `vncpasswd`

Fixes https://github.com/makandra/geordi/issues/124

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
     - name: Install vncserver
       run: |
         sudo apt-get update
-        sudo apt-get install -y vnc4server
+        sudo apt-get install -y tightvncserver
         # Fake required password for vncserver
         mkdir ~/.vnc && touch ~/.vnc/passwd && chmod 700 ~/.vnc/passwd
     - name: Bundle
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ## Unreleased
 
 ### Compatible changes
+* Switch to tightvncserver to be compatible with Ubuntu 20.04. vnc4server is still supported.
+
 ### Breaking changes
 
 
--- a/lib/geordi/commands/_setup_vnc.rb
+++ b/lib/geordi/commands/_setup_vnc.rb
@@ -23,25 +23,23 @@ def _setup_vnc
     Interaction.success 'It appears you already have a VNC server installed. Good job!'
   else
     puts 'Please run:'
-    Interaction.note_cmd 'sudo apt-get install vnc4server'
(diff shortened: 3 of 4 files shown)


makmic:
The PR seems to be red in Ruby 2.0 because of other reasons (same issue was already present in the last PR)

codener (approved):
Code looks fine. Please make sure the new approach works with both VNC solutions.

makmic:
> Code looks fine. Please make sure the new approach works with both VNC solutions.

Thanks for the review.
Could you verify the vnc4server-compatibility yourself? As I'm now on 20.04 I can no longer test it..

codener:
Installing Geordi with your changes successfully runs my tests without requiring any modifications :heavy_check_mark: 
Removing VNC viewer and server, then re-setting up with `geordi vnc --setup` worked as well and tests still run :heavy_check_mark: 

One detail however: Geordi's VNC setup still includes instructions for setting a secure password. Following the instructions, I was not asked for a password. I assume your changes make manual configuration obsolete? That part of the instructions should then be removed.

However, the instructions state that "if you start a vncserver without our cucumber script, a user with your password can connect to your machine". This sound like having a hard-coded public password is not a good idea. Or am I missing something?

makmic:
Thanks for verifying the new setup process!

I assume you were not asked for a default VNC password as vnc4server was previously installed (leaving the file `~/.vnc/passwd` behind). The password instructions are only required for vnc4server, not for tightvncserver as the latter can no longer be booted without a password - neither from geordi nor otherwise.

I'll adjust the installation instructions to differentiate between those two. Secondly I'll replace the password file with an auto generated file (not checked into git / custom for every user).

makmic:
@codener I think I found a cleaner aproach. Both servers (vnc4 and tightvnc) are now using the default file `~/.vnc/passwd` which seems to be the "industry standard". Please check again if you are OK with the code changes.

codener (approved):
Code looks fine, will perform another local test.

codener:
Verified both scenarios (Geordi update, initial vnc setup) are working. :heavy_check_mark: 

I'd like to avoid the ambiguity in the VNC server setup. Could we suggest to install `tightvncserver` and add a note to try `vnc4server` if the former is not available?
```
# Setup VNC server
Please run:
(pink)> sudo apt install tightvncserver
(normal)> In case this package is not available, you may try vnc4server instead.
Continue ...
```

makmic:
I just noticed a bug that somehow made it through to the release: `File.absolute_path('~/.vnc/passwd').freeze` does not resolve the path correctly, it needs to be replaced with `File.expand_path`. 
Sorry for the inconvenience, I'll open a new PR including a patchlevel version bump.

Issue #126, opened 2020-10-28
Remove Ruby Version 2.0


denzelem:
We want  to drop support for Ruby 2.0 in the next major release. Otherwise we would need to spend extra time in fixing the issues in the tests with Ruby 2.0.

denzelem:
Closed with #132.

PR #131, opened 2020-11-24
geordi auf Github Actions umbauen

niklas-hasselmeyer:


niklas-hasselmeyer:
Until vnc works with Ubuntu 20.04, the tests have to run on 18.04.

denzelem (approved):
Thanks, LGTM. Since we are dropping support for Ruby < 2.2 in the tests we should implement #126 as soon as possible. I will take care of it.

PR #139, opened 2021-01-19
Remove auto-bundling from geordi shell and geordi console

### LINKED ISSUE #136: Remove auto-bundling from `geordi shell` and `geordi console`
We're maintaining old projects inside a Docker container because we cannot install old Rubies on modern Ubuntus.

We want to keep using `geordi shell` and `geordi console`, but these commands don't work in a Docker container that have no SSH agent.

We could use geordi outside the docker container, but the shell and console commands automatically bundle. This will fail because there is no Ruby to bundle into outside Docker.

It would be good to remove auto-bundling, at least for these two commands.

Since this affects old projects in particular this change would need to be backported to the last Geordi version that supports Ruby 1.8.7.

niklas-hasselmeyer:
`geordi console` will still bundle if it is executed locally.
Gem version detection now works without prior bundling.
Closes #136

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Compatible changes
 
+* Remove auto-bundling from `geordi shell` and remote `geordi console`
+
 ### Breaking changes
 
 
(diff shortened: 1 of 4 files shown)


codener (approved):
Code has even become better by this. Thanks.

PR #140, opened 2021-01-29
Bs/Extend testing

brunosedler:


features/deploy.feature
```diff
@@ -5,24 +5,16 @@ Feature: The deploy command
...
 
-    When I run `geordi deploy` interactively
-      # Answer three prompts
-      And I type "staging"
-      And I type "master"
-      And I type ""
-      # Confirm deployment
-      And I type "yes"
+    When I run `geordi deploy` in interactive mode
+    Then I should see a prompt "Deployment stage: [staging]"
```
brunosedler, line 9: These changes would come at a high cost. Unfortunately, the command does not seem to write to stdout as long as it is still running. Therefore, to make this work, we'd have to stop the command, and start it again from beginning to make this work. See the step definitions below. This Feature alone in this state would take a minute to run, so I think this should be rolled back, but I wanted to show it and see what you think

features/step_definitions/aruba_steps.rb
```diff
@@ -0,0 +1,65 @@
...
+
+When(/^I type "([^"]*)" and continue$/) do |input|
+  @interactive.start
+  @processed_inputs << input
+  @processed_inputs.each do |step_input|
+    type(unescape_text(step_input))
+  end
+end
+
+Then /^the last (\d) lines of (output|stderr|stdout) should contain: "(.+?)"$/ do |line_count, channel, expected_output|
```
brunosedler, line 27: As mentioned above, we only see the output of a command after it has finished, so we cannot return the output from each step of an interactive command. 
With this step we now have the following options:
Aruba: "the (output|stderr|stdout) should contain" --> all output
Aruba: "the (output|stderr|stdout) from "command" should contain" --> output of a specific command
Custom: "The last n lines of (output|stderr|stdout) should contain" --> last lines of all output

I'm not sure if this is what you imagined. If so, i could add another step for checking the last lines of a specific command

spec/capistrano_config_spec.rb
```diff
@@ -0,0 +1,127 @@
+describe Geordi::CapistranoConfig, type: :aruba do
+  subject do
+    Geordi::CapistranoConfig.new('staging', Dir.getwd + '/tmp/aruba')
+  end
+
+  describe '#load_deploy_info' do
+    before { write_file('Capfile', '# Configure your capistrano tasks') }
```
codener, line 7: Instead of passing a custom root dir to CapistranoConfig, couldn't you set `ENV['RAILS_ROOT']` and have CapistranoConfig find its root itself? This would walk existing code instead of adding a custom workaround for tests only.

features/deploy.feature
codener: Thanks for the proof of concept, but I agree this is too high an overhead. Please revert.

features/step_definitions/aruba_steps.rb
codener: Having to know how many lines of output to ignore is a low-level detail. While this could help to solve the "accidentally matched output of a previous command", a step to reset previous output would be most helpful. Something like `Given I ignore previous output`. Could this be added instead?

features/step_definitions/aruba_steps.rb
```diff
@@ -0,0 +1,65 @@
...
+  output = case channel
+  when 'output'
+    all_output
+  when 'stderr'
+    all_stderr
+  when 'stdout'
+    all_stdout
+  end
```
codener, line 35: While reading this, I am asking myself: Are `all_output` & co Aruba methods? When they're called `all_*`, are there "partial *" counterparts?

spec/capistrano_config_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+      deploy_info = subject.send(:load_deploy_info)
+
+      expect(deploy_info).to eq(<<-TEXT)
+        set :rails_env, 'staging'
+        set :deploy_to, '/var/www/example.com'
+        set :user, 'user'
+
+        server 'www.example.com'
+
+        set :deploy_to, 'var/www.foobar.com'
```
codener, line 36: Since this test is nothing about file contents and only about concatenation, you could make it more readable by using self-explanatory file contents. The expected deploy info could then be:

```
content of config/deploy/staging.rb
content of config/deploy.rb
```

codener (commented):
Thanks for adding RSpec and writing a first spec, and thanks for your research. Please see the code comments.

One general remark: commit messages are usually put in present tense (Indikativ Präsens). Please consider this when squashing your commits.

features/step_definitions/aruba_steps.rb
brunosedler: Yes they are aruba methods. `all_output` in this context means, the output of all commands that have been run in this scenario. The only "patial" output given by aruba therefore is the output of a specific command, which can be accessed  by the step  `Then(/^(?:the )?(output|stderr|stdout) from "([^"]*)" should( not)? contain( exactly)? "([^"]*)"$/)`. But there are no `partial_*` or similar counterparts

features/step_definitions/aruba_steps.rb
brunosedler: If you only want to make sure that you do not accidentally match the output of another command, you could use the step `Then(/^(?:the )?(output|stderr|stdout) from "([^"]*)" should( not)? contain( exactly)? "([^"]*)"$/)`.  That only checks the output of the command you specify in the `from "([^"]*)"` part.

I thought, this step here could be useful if you don't want to check for whole text passages or sentences but only specific words or phrases, which may appear multiple times in an output, but you only want to see if they appear somewhere near the end, where they may be more relevant.

spec/spec_helper.rb
```diff
@@ -0,0 +1,103 @@
...
+  # the seed, which is printed after each run.
+  #     --seed 1234
+  config.order = :random
+
+  # Seed global randomization in this process using the `--seed` CLI option.
+  # Setting this allows you to use `--seed` to deterministically reproduce
+  # test failures related to randomization by passing the same `--seed` value
+  # as the one that triggered the failure.
+  Kernel.srand config.seed
+=end
```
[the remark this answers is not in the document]
brunosedler: I agree, i don't think we need the profiling for each run. I think that would be something you enable when you specifically try to improve existing tests, but else it's just unnecessary extra text.

But enabling the rest makes sense :+1:

features/step_definitions/aruba_steps.rb
brunosedler: But to also answer your question, yes it would be possible to reset the previous output. This would be achieved by resetting the `command_monitor` that holds all commands that have been run so far.

features/step_definitions/aruba_steps.rb
codener: Ok thanks.

features/step_definitions/aruba_steps.rb
codener: Thanks for hinting that step, might be helpful. However, I still like an explicit `I ignore previous output`. Would you add that and remove all code related to "n lines" (including the helper at the bottom)?

codener (approved):
Thanks for putting effort into this. Sorry to have reverted of your work. However, this should be a good state now.

PR #141, opened 2021-01-29
Allow pivotal tracker projects ids in global config file

martinschaflitzl1:
Some pivotal tracker projects contain stories for multiply projects (e.g. WeeklyWombat). When working on theses stories it would be nice to use `geordi commit` in all projects without having to edit the local config every time.

PR #144, opened 2021-02-16
Allow `geordi commit` to continue if one of the projects is inaccessible

niklas-hasselmeyer:
As discussed [here](https://github.com/makandra/geordi/pull/141#issuecomment-776096374).

### DIFF
--- a/features/commit.feature
+++ b/features/commit.feature
@@ -2,6 +2,14 @@ Feature: Creating a git commit from a Pivotal Tracker story
   Background:
     Given a file named "tmp/global_settings.yml" with "pivotal_tracker_api_key: my_api_key"
 
+
+  Scenario: If there are no stories, the command fails
+    Given there are no stories
+
+    When I run `geordi commit`
+    Then the output should contain "No stories to offer."
+
+
   Scenario: Extra arguments are forwarded to "git commit"
     Given I have staged changes
 
--- a/lib/geordi/gitpt.rb
+++ b/lib/geordi/gitpt.rb
@@ -64,10 +64,21 @@ No staged changes. Will create an empty commit.
 
     def load_projects
       project_ids = settings.pivotal_tracker_project_ids
-      project_ids.collect { |project_id| client.project(project_id) }
+      project_ids.collect do |project_id|
+        begin
+          client.project(project_id)
+        rescue TrackerApi::Errors::ClientError
+          puts # Start a new line
+          Geordi::Interaction.warn "Could not access project #{project_id}. Skipping."
+        end
+      end.compact
     end
 
     def applicable_stories
+      if Util.testing?
+        return ENV['GEORDI_TESTING_NO_PT_STORIES'] == 'true' ? [] : [OpenStruct.new(id: 12, name: 'Test Story')]
+      end
+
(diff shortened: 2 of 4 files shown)


lib/geordi/gitpt.rb
```diff
@@ -54,6 +61,8 @@ def choose_story
     stories = applicable_stories
     reset_loading_message = "\r#{' ' * (loading_message.length + stories.length)}\r"
 
+    Geordi::Interaction.fail('No stories found') if stories.empty?
```
codener, line 64: Please fail with a sentence, and change wording:
```suggestion
    Geordi::Interaction.fail('No stories to offer.') if stories.empty?
```

lib/geordi/gitpt.rb
```diff
@@ -34,7 +34,14 @@ def build_client
 
   def load_projects
     project_ids = settings.pivotal_tracker_project_ids
-    project_ids.collect { |project_id| client.project(project_id) }
+    project_ids.collect do |project_id|
+      begin
+        client.project(project_id)
+      rescue TrackerApi::Errors::ClientError
+        puts # Start a new line
+        Geordi::Interaction.warn "Could not load project #{project_id}. Skipped."
```
codener, line 42: Please use this:
```suggestion
        Geordi::Interaction.warn "Could not access project #{project_id}. Skipping."
```

codener (commented):
Thanks, this is a good addition. I have made a few adjustments in wording.

Please add tests: one for skipping invalid project ids, and one for seeing an error if no stories can be offered.

niklas-hasselmeyer:
A test for skipping invalid project ids is close to impossible without actually querying the PT API and hoping that it returns an error, so I didn't add a test for that.

features/commit.feature
```diff
@@ -2,6 +2,14 @@ Feature: Creating a git commit from a Pivotal Tracker story
   Background:
     Given a file named "tmp/global_settings.yml" with "pivotal_tracker_api_key: my_api_key"
 
+
+  Scenario: If there are no stories, the command fails
+    Given there are no stories
+
+    When I run `geordi commit` interactively
```
codener, line 9: You only need "interactively" when you intend to interact with the CLI. In this case you should omit it in order to not confuse your fellow developers.

lib/geordi/gitpt.rb
```diff
@@ -69,29 +69,33 @@ def load_projects
...
-        return OpenStruct.new(id: 12, name: 'Test Story')
-      end
-
       loading_message = 'Connecting to Pivotal Tracker ...'
       print(loading_message)
       stories = applicable_stories
       reset_loading_message = "\r#{' ' * (loading_message.length + stories.length)}\r"
 
-      Geordi::Interaction.fail('No stories found') if stories.empty?
+      Geordi::Interaction.fail('No stories to offer.') if stories.empty?
```
codener, line 94: Remember that you do not need the `Geordi::` prefix when you're inside the `Geordi` namespace (like with `Geordi::Util` below).

codener (approved):
LGTM! Thank you for the good bash UX.

Before merging, please remove the two hidden `.swp` files.

PR #153, opened 2021-04-06
Remove VNC and Firefox support for integration tests

### LINKED ISSUE #151: Remove VNC support in geordi 6
* VNC will not be supported in geordi 6 anymore
* All projects using geordi should try to use a headless Chrome https://makandracards.com/makandra/492109-capybara-running-tests-with-headless-chrome
* For legacy projects you still might want to continue with the usage of geordi 5 or geordi v2.12.3 (last version with Ruby 1.8.7 support)
* Add a deprecation warning if the `use_vnc: false` setting is present in `.geordi.yml`
* Remove all VNC code from geordi

steevokovsky-mkd:
With Ubuntu 20.04, the VNC server causes problems and we therefore don't want to use VNC for projects outside Docker. 
Hence, we remove the whole VNC/Firefox logic.

Closes #151


features/docker.feature
```diff
@@ -84,3 +84,14 @@ Feature: The docker command
...
+
+  Scenario: A VNC viewer can connect to the docker container
+    Given a file named "docker-compose.yml" with:
+    """
+    services:
+      main: foo
+    """
+
+    When I run `geordi docker vnc`
+    Then the output should contain ""
```
denzelem, line 97: Missing output

lib/geordi/commands/_setup_vnc.rb
```diff
@@ -2,47 +2,19 @@
...
   `clear`
 
-  Interaction.note 'This script will help you install a VNC server and a VNC viewer.'
+  Interaction.note 'This script will help you install a VNC viewer.'
   puts
   puts Util.strip_heredoc <<-TEXT
-    With those you will be able to use our cucumber script without being
-    disturbed by focus-stealing Selenium windows. Instead, they will open
-    inside a VNC session.
+    With that you will be able to use our docker vnc script to open
```
denzelem, line 8: `...to use our docker vnc script to open` -> `...to use our docker VNC script to open`

lib/geordi/settings.rb
```diff
@@ -81,6 +74,11 @@ def read_settings
       if File.exists?(local_path)
         local_settings = YAML.safe_load(File.read(local_path))
         check_for_invalid_keys(local_settings, ALLOWED_LOCAL_SETTINGS, local_path)
+
+      if File.exists?(File.expand_path("../../.firefox-version"))
+        Geordi::Interaction.warn 'Unsupported setting ".firefox-version". Please remove the file in /home/user/projects/the-project/.firefox-verison.'
```
denzelem, line 79: Can you please replace the path with the path from `File.expand_path("../../.firefox-version")`? So the user knows where she needs to delete the file.

lib/geordi/settings.rb
```diff
@@ -93,7 +91,11 @@ def check_for_invalid_keys(settings, allowed_keys, file)
         Geordi::Interaction.warn "Geordi detected unknown keys in #{file}.\n"
 
         invalid_keys.sort.each do |key|
-          puts "* #{key}"
+          if key == :use_vnc
+            puts '* use_vnc: Unsupported setting "use_vnc". Please remove this setting from /home/user/projects/the-project/.geordi.yml.'
```
denzelem, line 95: Please replace the path with the real path of the settings, so the user knows where she needs to delete the file.

denzelem (commented):
Can you please create a second commit, with the version bump (compare with ae2afef368a6bcc188b1fea42442c06238d189ec)? Then I can merge and release the changes once you adjusted the pending threads.

lib/geordi/commands/_setup_vnc.rb
```diff
@@ -56,8 +28,8 @@ def _setup_vnc
 
   puts
   puts Util.strip_heredoc <<-TEXT
-    All done. Our cucumber script will now automatically run Selenium features
-    in VNC.
+    All done. Our docker vnc script will now automatically open a VNC viewer 
+    to connect to the docker container.
```
codener, line 32:
```suggestion
    Done. You can view the VNC window with `geordi docker vnc`.
```

lib/geordi/commands/_setup_vnc.rb
```diff
@@ -2,47 +2,19 @@
...
-    disturbed by focus-stealing Selenium windows. Instead, they will open
-    inside a VNC session.
+    With that you will be able to use our docker VNC script to open
+    a VNC viewer for the current docker container.
 
     You can still inspect everything with:
   TEXT
-  Interaction.note_cmd 'geordi vnc'
+  Interaction.note_cmd 'geordi docker vnc'
   puts
```
codener, line 14: Please remove altogether. This only had relevance for new VNC users.

lib/geordi/cucumber.rb
```diff
@@ -4,69 +4,33 @@
...
-      end
+      system command
     end
 
     def restore_env
-      VNC_ENV_VARIABLES.each do |variable|
+      ENV_VARIABLES.each do |variable|
         ENV[variable] = ENV["OUTER_#{variable}"]
       end
     end
```
codener, line 32: This must be removed, as you're not setting the OUTER_* vars any more. Please do an extra check for similar usage.

lib/geordi/commands/cucumber.rb
```diff
@@ -17,9 +16,7 @@
 e.g. `--format=pretty`. Make sure to connect option and value with an equals
 sign, i.e. have each option a contiguous string.
 
-- *VNC:* By default, test browsers will run in a VNC session. When using a
-headless test browser anyway, you can disable VNC by setting `use_vnc: false`
-in `.geordi.yml` in the project root.
+- By default, test browsers will run in a headless Chrome session.
```
codener, line 19: Remove this item altogether.

lib/geordi/commands/cucumber.rb
```diff
@@ -2,9 +2,8 @@
 long_desc <<-LONGDESC
 Example: `geordi cucumber features/authentication_feature:3`
 
-Runs Cucumber with `bundle exec`, using parallel tests, with a VNC session
-holding Selenium test browsers, support for using a dedicated testing browser
-and beta support for re-running failed scenarios.
+Runs Cucumber with `bundle exec`, using parallel tests, with a headless Chrome session, 
+with beta support for re-running failed scenarios.
```
codener, line 6: Please do not mention Headless Chrome, as it is a project configuration thing. Geordi fortunately does not have to know about it.

```suggestion
Runs Cucumber with `bundle exec`, using parallel tests and with support for re-running
failed scenarios.
```

README.md
```diff
@@ -90,9 +90,8 @@ Run Cucumber features.
 
 Example: `geordi cucumber features/authentication_feature:3`
 
-Runs Cucumber with `bundle exec`, using parallel tests, with a VNC session
-holding Selenium test browsers, support for using a dedicated testing browser
-and beta support for re-running failed scenarios.
+Runs Cucumber with `bundle exec`, using parallel tests, with a headless Chrome session, 
+with beta support for re-running failed scenarios.
```
codener, line 94: Please remember to update the README when updating the command description.

lib/geordi/cucumber.rb
```diff
@@ -4,69 +4,33 @@
...
 
 module Geordi
   class Cucumber
 
-    VNC_DISPLAY = ':17'.freeze
-    VNC_PASSWORD_FILE = File.expand_path('~/.vnc/passwd').freeze # default for "vncpasswd"
-    VNC_SERVER_DEFAULT_OPTIONS = "-localhost -nolisten tcp -geometry 1280x1024 -rfbauth #{VNC_PASSWORD_FILE}".freeze
-    VNC_SERVER_COMMAND = "vncserver #{VNC_DISPLAY} #{ENV.fetch('GEORDI_VNC_OPTIONS', VNC_SERVER_DEFAULT_OPTIONS)}".freeze
-    VNC_VIEWER_COMMAND = "vncviewer -passwd #{VNC_PASSWORD_FILE}".freeze
-    VNC_ENV_VARIABLES = %w[DISPLAY BROWSER LAUNCHY_BROWSER].freeze
```
codener, line 18: Where did all the logic for running tests inside VNC go? Was it ported to our legacy project docker setup?

lib/geordi/docker.rb
```diff
@@ -5,6 +5,9 @@
 module Geordi
   class Docker
     DOCKER_COMPOSE_FILE = 'docker-compose.yml'.freeze
+    VNC_DISPLAY = ':17'.freeze
+    VNC_PASSWORD_FILE = File.expand_path('~/.vnc/passwd').freeze # default for "vncpasswd"
```
codener, line 9: How can this work when you removed the password setting process in _setup_vnc.rb?

lib/geordi/docker.rb
```diff
@@ -5,6 +5,9 @@
 module Geordi
   class Docker
     DOCKER_COMPOSE_FILE = 'docker-compose.yml'.freeze
+    VNC_DISPLAY = ':17'.freeze
```
codener, line 8: When L35 always invokes `launch_vnc_viewer` with an argument, how is this constant still needed? Should be removed IMHO.

lib/geordi/docker.rb
```diff
@@ -28,11 +31,31 @@ def shell(options = {})
...
+        end
+        unless $?.success?
+          if $?.exitstatus == 127
+            fail('VNC viewer not found. Install it with `geordi docker vnc --setup`.')
+          else
+            fail("VNC viewer could not be opened: #{error}")
+          end
+        end
+      end
+      success('Launching VNC viewer was successful.')
```
codener, line 56: `success` is intended as a conclusion of an action that was `announce`d. Since the opened VNC window proves success, please simply remove this line.

codener (changes requested):
Thanks for the effort and thorough work. Have a bunch of remarks tho. Please have a look.

lib/geordi/docker.rb
steevokovsky-mkd: In order to test the `geordi docker vnc` command, I added this success message. The output would not contain any feedback without this, but I can replace that with a puts if that is more fitting?

lib/geordi/cucumber.rb
```diff
@@ -4,69 +4,33 @@
...
-          if $?.exitstatus == 127
-            Interaction.fail 'VNC viewer not found. Install it with `geordi vnc --setup`.'
-          else
-            Interaction.note 'VNC viewer could not be opened:'
-            puts error
-            puts
-          end
-        end
-      end
+      system command
```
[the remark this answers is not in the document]
steevokovsky-mkd: The `Util.run!` results in features failing. The actual output seems to change when using system or Util.run!. The features expect the output received with the system command, that's why I used it. Here is an example of a feature failing with Util.run! and succeeding with system (features/cucumber.feature:141, Using the @solo tag):

```
...
When I run `geordi cucumber --verbose features`
Then the output should contain "# Running @solo features"
  And the output should match /^> .*cucumber .*--tags @solo/
   And the output should contain "# Running features"  <------- failing here
   And the output should match /^> .*cucumber .*--tags \"not @solo\"/
```
I'd change the old comment `# Util.run! would reset the Firefox PATH` to something like `# Using Util.run! would result in missing output`. Would you recommend something else?

lib/geordi/docker.rb
codener: This should help you without a success message: https://relishapp.com/cucumber/aruba/v/0-11-0/docs/command/check-exit-status-of-commands

lib/geordi/cucumber.rb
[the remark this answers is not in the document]
codener: I see. Probably the code that invokes this method relies on its success returned as Boolean, and `Util.run!` does not return it. As we're not 100% sure what makes the test fail, please use `system` without a comment.

lib/geordi/docker.rb
steevokovsky-mkd: If someone already used VNC, there is already a passwd file present in the users root directory. It also depends on the authentication method if this is even needed, after some testing it seems like the server (in this case the VNC server running in the docker container), defines the used passwd file at initialization. As far as the docs go, that password file, given it does not change during runtime, does this:
> The passwd-file specifies the file containing the password used to authenticate viewers if VNC authentication is in use. 

Further testing will clear this up I hope.

lib/geordi/cucumber.rb
steevokovsky-mkd: The legacy docker is running with a old geordi version, so running tests inside VNC is handled by two instances really:
- The docker container is starting a VNC server with a passwd file located in the containers root directory.
- This is triggered by running tests from the containers shell (e.g via `geordi cucumber feature/feature_xy.feature` 
- The VNC viewer can be started via `geordi docker vnc` from the local shell, opening a VNC window connected to the VNC server in the docker container.

lib/geordi/docker.rb
steevokovsky-mkd: It turns out that the password option was never really used. Though the user can provide the password file, the docker-compose.yml in our legacy docker setup disabled this with `GEORDI_VNC_OPTIONS: '-SecurityTypes None -geometry 1280x1024'`

Using docker with a VNC viewer seems to mitigate security issues. I think this flag was added by @kratob, but I don't know for sure.

I hope this answeres your question.

codener (approved):
Schaut gut aus soweit. Ob da jetzt alles richtig ist und funktioniert, kann ich vom Code her nicht beurteilen.

Bitte merge in master, ich schau dass ein paar Leute das lokal ausprobieren.

PR #158, opened 2021-04-29
@solo tag aus geordi entfernen

### LINKED ISSUE #155: Remove @solo support
Geordi's `cucumber` command runs scenarios tagged with @solo serially for themselves. This was introduced to handle focus issues in VNC, where only one of the test browsers could have focus. With the transition from VNC to Headless Chrome, this is not required any more.

- Remove support for @solo tag
- Streamline `cucumber` command accordingly
- Notify users via `spec.post_install_message`

steevokovsky-mkd:
Closes #155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ## Unreleased
 
 ### Compatible changes
+* Removed support for serial execution of scenarios tagged with @solo.
 
 ### Breaking changes
 
--- a/features/cucumber.feature
+++ b/features/cucumber.feature
@@ -19,7 +19,7 @@ Feature: The cucumber command
     Then the output should contain "# Running features"
       And the output should contain "> Only: features/single.feature"
       And the output should contain "Features green."
-    But the output should not contain "parallel"
+    But the output should not contain "> Using parallel_tests"
 
 
   Scenario: Multiple features are run in parallel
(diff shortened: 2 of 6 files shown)


features/cucumber.feature
```diff
@@ -138,11 +138,7 @@ Feature: The cucumber command
       And the output should contain "> Using parallel_tests"
 
 
-  Scenario: Using the @solo tag
-
-    The cucumber command runs all features without the @solo tag. If any of the
-    specified .feature files contains '@solo', it boots Cucumber a second time
-    and runs only those of the features tagged with @solo.
+  Scenario: The usage of @solo tags is ignored
```
brunosedler, line 141: Drop this scenario as well. The `@solo` tag is no longer part of geordi's functionality, so we don't need any tests concerning `@solo` anymore.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ## Unreleased
 
 ### Compatible changes
+* Removed support for serial execution of scenarios tagged with @solo.
```
brunosedler, line 9: Not sure if this might not be a breaking change :thinking: 
It does not break code but it is an unexpected change in behavior. What do you think @codener

codener: It's tricky. Since serial execution was only ever needed for parallel tests inside VNC, and VNC support is already removed, I'd call it "compatible change". On the other hand, users tagging scenarios with `@solo` must adapt. However, obsolete `@solo` tags do no harm. Let's go with "compatible".

brunosedler (approved):

PR #159, opened 2021-05-06
[#177576666] Avoid exiting after invalid config keys are detected

steevokovsky-mkd:


features/cucumber.feature
```diff
@@ -256,3 +256,16 @@ Feature: The cucumber command
     # Passing a line number activates serial test execution
     When I run `geordi cucumber --verbose single.feature:1`
     Then the output should contain "> bin/cucumber"
+
+
+  Scenario: Invalid keys are reported once per settings file
```
codener, line 261: The scenario title sounds like it should ensure each settings file prints warnings, which is not the case. Please change to:

```suggestion
  Scenario: Invalid config keys are not reported twice
```

features/step_definitions/aruba_steps.rb
```diff
@@ -6,3 +6,13 @@
 Given 'I ignore previous output' do
   aruba.command_monitor.clear
 end
+
+Then(/^(?:the )?output should( not)? contain "([^"]*)" once$/) do |negated, expected|
```
codener, line 10: Please do not add support for "not". Firstly, you don't need it. Secondly, it is ambiguous. `The output should not contain "hello" once` sounds like "should not occur anywhere", but your implementation would also be green with two or more occurrences.

lib/geordi/settings.rb
```diff
@@ -78,36 +78,22 @@ def read_settings
...
-          exit 1
-        end
+      return if ENV['GEORDI_INVALID_SETTINGS_WARNED'].to_s.include? file
+      return if settings.empty?
+
+      invalid_keys = settings.keys - allowed_keys
+      unless invalid_keys.empty?
+        Interaction.warn "Unknown settings in #{file}: #{invalid_keys.join(", ")}"
+        Interaction.note "Supported settings in #{file} are: #{allowed_keys.join(", ")}"
+        ENV['GEORDI_INVALID_SETTINGS_WARNED'] = "#{ENV['GEORDI_INVALID_SETTINGS_WARNED']} #{file}"
```
codener, line 96: Right, we did not consider that there are multiple config files. Your fix will probably work, but it feels hacky. Also, you did not add a test (should have been a spec). I believe, it shows that the "did we already warn" check is placed at the wrong location. So please

* Keep the simplification of this method (`return if settings.empty?`)
* Extract L70-82 to `validated_configuration`
* In this new method, do the check with a boolean `ENV['GEORDI_INVALID_SETTINGS_WARNED']`

codener (changes requested):
Thanks for the update. I believe I found a better solution, please have a look and try it out.

lib/geordi/settings.rb
steevokovsky-mkd: I don't see the benefit of extracting L70-82 into `validated_configuration`. Wouldn't it be possible to check for the boolean in the read_settings function? That way the paths above (global/local/firefox_path) are in the correct scope:

```
def read_settings
    global_path = GLOBAL_SETTINGS_FILE_NAME
    local_path = LOCAL_SETTINGS_FILE_NAME
    firefox_path = File.expand_path("../../.firefox-version")

    unless ENV['GEORDI_INVALID_SETTINGS_WARNED']
        if File.exists?(global_path)
            global_settings = YAML.safe_load(File.read(global_path)) || {}
            check_for_invalid_keys(global_settings, ALLOWED_GLOBAL_SETTINGS, global_path)
        end

        if File.exists?(local_path)
            local_settings = YAML.safe_load(File.read(local_path)) || {}
            check_for_invalid_keys(local_settings, ALLOWED_LOCAL_SETTINGS, local_path)
        end
    end

    Interaction.warn "Unsupported config file \".firefox-version\". Please remove #{firefox_path}." if File.exists?(firefox_path)

    @global_settings = global_settings || {}
    @local_settings = local_settings || {}
end
```

After that, we just set ENV['GEORDI_INVALID_SETTINGS_WARNED']  to 'true' whenever we warn the user about the settings.

codener: Right, this is similar to what I asked you. However, the solution I suggested has three advantages:

* Extracting the settings validation into its own method improves code structure and readability
* Including the "firefox version" check in that function ensures the warning will also printed once at most, which proves this is the right location to set and read the ENV variable
* It narrows the scope of that ENV variable, because you can simply `return` at the beginning of that method if it is set, and set it at the end of the method. Other developers have less to wonder about that variable.

Please also extend the Cucumber test for this feature so it ensures a consecutive execution of Geordi (i.e. the cucumber command) will warn again, i.e. consecutive warnings are only suppressed for the current execution, and this is not persisted.

features/cucumber.feature
```diff
@@ -256,3 +256,35 @@ Feature: The cucumber command
...
+    Feature: Running multiple features
+      Scenario: A scenario
+      Scenario: Another scenario
+    """
+
+    When I run `geordi cucumber features/cucumber.feature`
+    Then the output should contain '> Unknown settings in ./tmp/global_settings.yml: use_vnc' 1 time
+      And the output should contain '> Supported settings in ./tmp/global_settings.yml are: ' 1 time
+
+    When I run `geordi cucumber features/single.feature`
```
codener, line 288: As far as I can see, `single.feature` does not exist. Although tests are green, please prefer to run the existing `cucumber.feature` here.

features/step_definitions/aruba_steps.rb
```diff
@@ -6,3 +6,8 @@
 Given 'I ignore previous output' do
   aruba.command_monitor.clear
 end
+
+Then(/^(?:the )?output should contain '([^']*)' (\d) times?$/) do |expected,amount|
+  console_output = aruba.command_monitor.all_output
+  expect(console_output.scan(expected).length).to eq amount.to_i
+end
```
codener, line 13: :ok_hand:

features/cucumber.feature
```diff
@@ -256,3 +256,35 @@ Feature: The cucumber command
...
+    Then the output should contain '> Unknown settings in ./tmp/global_settings.yml: use_vnc' 2 times
```
codener, line 289: In order to avoid misinterpretations, please use the `I ignore previous output` step before L288, and then expect the output again "once".

lib/geordi/settings.rb
```diff
@@ -67,47 +67,34 @@ def read_settings
...
+      local_settings = File.exists?(local_path) ? YAML.safe_load(File.read(local_path)) || {} : {}
+      check_for_invalid_keys(local_settings, ALLOWED_LOCAL_SETTINGS, local_path)
+
+      Interaction.warn "Unsupported config file \".firefox-version\". Please remove #{firefox_path}." if File.exists?(firefox_path)
+
+      ENV['GEORDI_INVALID_SETTINGS_WARNED'] = 'true'
+      [global_settings, local_settings]
+    end
+
     def check_for_invalid_keys(settings = [], allowed_keys, file)
```
codener, line 91: Can you remove the default value? Since `settings` usually is a Hash, the Array is misleading. Also, a leading default value behaves unintuitive, as `check_for_invalid_keys([:a], [:b])` would result in arguments `[], [:a], [:b]`.

lib/geordi/settings.rb
```diff
@@ -67,47 +67,34 @@ def read_settings
...
+      check_for_invalid_keys(global_settings, ALLOWED_GLOBAL_SETTINGS, global_path)
+
+      local_settings = File.exists?(local_path) ? YAML.safe_load(File.read(local_path)) || {} : {}
+      check_for_invalid_keys(local_settings, ALLOWED_LOCAL_SETTINGS, local_path)
+
+      Interaction.warn "Unsupported config file \".firefox-version\". Please remove #{firefox_path}." if File.exists?(firefox_path)
+
+      ENV['GEORDI_INVALID_SETTINGS_WARNED'] = 'true'
+      [global_settings, local_settings]
+    end
```
codener, line 89: Thanks for refactoring this. However, I apologize: I did overlook that loading the settings was done in L71 and L76. Moving the loading to `validate_configuration` does not make much sense.

So, lets elaborate on your last suggestion:

```ruby
SETTINGS_WARNED = 'GEORDI_INVALID_SETTINGS_WARNED'
# ...

def read_settings
  global_path = GLOBAL_SETTINGS_FILE_NAME
  local_path = LOCAL_SETTINGS_FILE_NAME

  global_settings = if File.exists?(global_path)
    YAML.safe_load File.read(global_path)
  end

  local_settings = if File.exists?(local_path)
    YAML.safe_load File.read(local_path)
  end

  # Prevent duplicate warnings caused by another instance of Settings
  unless ENV[SETTINGS_WARNED]
    check_for_invalid_keys(global_settings, ALLOWED_GLOBAL_SETTINGS, global_path)
    check_for_invalid_keys(local_settings, ALLOWED_LOCAL_SETTINGS, local_path)
  
    firefox_path = File.expand_path("../../.firefox-version")
    if File.exists?(firefox_path)
      Interaction.warn "Unsupported config file \".firefox-version\". Please remove #{firefox_path}."
    end

    ENV[SETTINGS_WARNED] = 'true'
  end

  @global_settings = global_settings || {}
  @local_settings = local_settings || {}
end
```

`check_for_invalid_keys` then needs to return if `settings` is `nil`.

Please extend your Cucumber scenario for the `.firefox-version` warning. I doubt the code is correct, since Geordi is mostly executed in project root. `File.expand_path('../../.firefox-version')` would not find that file in project root.

codener (changes requested):
Thanks for the update. However, I did not read the codes right and we should do another improvement.

features/cucumber.feature
```diff
@@ -256,3 +256,47 @@ Feature: The cucumber command
...
     Then the output should contain "> bin/cucumber"
+
+
+  Scenario: Invalid config keys are not reported twice in the same geordi command
+    Given a file named "tmp/global_settings.yml" with "use_vnc: false"
+    Given a file named "features/cucumber.feature" with:
+    """
+    Feature: Running multiple features
+      Scenario: A scenario
+      Scenario: Another scenario
```
codener, line 267: Bitte mach hier und an anderen Stellen einfach 

```suggestion
    Feature: A feature
      Scenario: A scenario
```
Dass es mehrere Scenarios sind, spielt keine Rolle, ebensowenig wie mehrere Features. Das verwirrt nur künftige Entwickler.

lib/geordi/settings.rb
```diff
@@ -11,6 +11,8 @@ class Settings
     ALLOWED_GLOBAL_SETTINGS = %w[ pivotal_tracker_api_key auto_update_chromedriver pivotal_tracker_project_ids ].freeze
     ALLOWED_LOCAL_SETTINGS = %w[ pivotal_tracker_project_ids ].freeze
 
+    SETTINGS_WARNED = 'false'
```
codener, line 14: Achtung, fieses Missverständnis: du verwendest diese Variable _als Namen für die ENV-Variable_. So heißt diese nun `false`. Bitte verwende wie zuvor `GEORDI_INVALID_SETTINGS_WARNED` oso.

codener (approved):
Schön geworden. Nur noch Details.

codener (approved):

PR #160, opened 2021-05-27
Removed spaces from post install message and clunky console output test.

steevokovsky-mkd:
Resolving issues regarding the post install message string and the cucumber.feature:
- The indentation of the post install message string currently prints two spaces verbatim to the terminal.
  - The two spaces should be removed.
- The "should not contain" test in `cucumber.feature:22` is satisfied accidentally if e.g. parallel tests are active but someone changed the output to `> With parallel_tests`.
  - They should make a generic expectation or be removed if obsolete.

### DIFF
--- a/features/cucumber.feature
+++ b/features/cucumber.feature
@@ -19,7 +19,7 @@ Feature: The cucumber command
     Then the output should contain "# Running features"
       And the output should contain "> Only: features/single.feature"
       And the output should contain "Features green."
-    But the output should not contain "> Using parallel_tests"
+    But the output should not contain "parallel"
 
 
   Scenario: Multiple features are run in parallel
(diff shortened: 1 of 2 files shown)


features/cucumber.feature
```diff
@@ -19,7 +19,6 @@ Feature: The cucumber command
     Then the output should contain "# Running features"
       And the output should contain "> Only: features/single.feature"
       And the output should contain "Features green."
-    But the output should not contain "> Using parallel_tests"
```
codener, line 22: I believe this cannot be removed. Please revert to the previous `should not contain "parallel"`.

codener (changes requested):

codener (approved):

PR #161, opened 2021-05-31
Allow `geordi tests` to take arguments

### LINKED ISSUE #152: Allow arguments for `geordi tests`
During local development I find myself running single specs and features in high frequency. Too often it then happens that I run `geordi rspec features/frontend/foo.feature`. The Rspec parser obviously can't parse the feature file and I have to remind myself to run the command again with `geordi cucumber` after wondering for a moment about the failing test.

I would love if `geordi test(s)` could capsulate the test-runner decision for me. So, I'd suggest:
* `geordi tests` still runs the entire test suite
* `geordi tests path/to/spec.rb` runs the test with Rspec (or TestUnit, whatever is bundled)
* `geordi tests path/to/test.feature` runs the test with Cucumber
* Bonus:  `geordi tests path/to/spec.rb path/to/test.feature` works as well

What do you think?

niklas-hasselmeyer:
Closes #152

### DIFF
--- a/features/tests.feature
+++ b/features/tests.feature
@@ -1,0 +1,52 @@
+Feature: The tests command
+
+  Scenario: Run all tests
+    Given an empty file named "test/test_helper.rb"
+      And an empty file named "spec/spec_helper.rb"
+      And an empty file named "features/some.feature"
+    When I run `geordi tests`
+    Then the output should contain "# Running Test::Unit"
+      And the output should contain "# Running specs"
+      And the output should contain "# Running features"
+
+  Scenario: Run only certain specs
+    Given an empty file named "spec/spec_helper.rb"
+      And an empty file named "spec/some_spec.rb"
+
+    When I run `geordi tests spec/some_spec.rb`
+    Then the output should contain "# Running specs"
+      And the output should contain "Only: spec/some_spec.rb"
+
+  Scenario: Run only certain features
+    Given an empty file named "features/some.feature"
+
+    When I run `geordi tests features/some.feature`
+    Then the output should contain "# Running features"
+      And the output should contain "> Only: features/some.feature"
+
+  Scenario: Run only tests in a directory
+    Given an empty file named "spec/spec_helper.rb"
+      And an empty file named "features/some.feature"
+
+    When I run `geordi tests spec`
+    Then the output should contain "# Running specs"
+      And the output should contain "Only: spec"
+
+    When I run `geordi tests features`
+    Then the output from "geordi tests features" should contain "# Running features"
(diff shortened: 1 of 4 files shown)


lib/geordi/commands/tests.rb
```diff
@@ -1,14 +1,40 @@
...
 
-  Interaction.success 'Successfully ran tests.'
+    if args.empty?
+      Interaction.fail "geordi tests only supports arguments if one or more test files/directories are specified."
+    elsif args.first.start_with? 'spec'
+      invoke 'rspec', args, opts
+    elsif args.first.start_with? 'features'
+      invoke 'cucumber', args, opts
+    else
+      Interaction.fail "The first argument #{args.first} could not be classified as either a rspec or a cucumber test."
```
brunosedler, line 24: I think this error message should be more precise. It hints that we try to parse file types which we don't. And I think we should also tell users how to use it correctly. Something like  
```
geordi tests cannot process the argument #{args.first}. The first argument has to be a path beginning with "features" or "spec". 
```

features/tests.feature
```diff
@@ -24,15 +24,31 @@ Feature: The tests command
...
+    Then the output should not contain "> bundle exec cucumber"
+
   Scenario: Unknown options are passed through
     Given an empty file named "spec/spec_helper.rb"
 
     When I run `geordi tests spec/ -c`
     Then the output should contain "Util.run! bundle exec rspec spec/ -c"
 
+    When I run `geordi tests spec/`
+    Then the output should not contain "Util.run! bundle exec rspec spec/ -c"
```
brunosedler, line 54: I think this doesn't make it any clearer. If I don't type `-c` I'd be pretty surprised to see it in the output. Is the Feature still green if the step is
```
Then the output should not contain "Util.run! bundle exec rspec spec/"
``` 
?
Otherwise, is `-c` a valid command line option for rspec? If so what does it do? i couldn't find any information about it.  
Maybe you can find an example where `geordi tests spec/ --some-option` produces an output that is different from running `geordi tests spec/ ` in other regards than the  `-some-option` missing.

niklas-hasselmeyer: `geordi rspec` doesn't do anything to the arguments, it just appends them if they are there. As rspec never actually runs in tests, the only difference in the output is the presence or absence of the option.

brunosedler: Okay. I expected aruba to actually run the rspec command and give us its output. But as it is, I think you can drop these two lines again. As I said before, if I don't type `-c` I'd be pretty surprised to see it in the output. Sorry for the back and forth, I just thought there's an easy way to make the scenario more conclusive.

brunosedler (approved):

PR #172, opened 2022-02-17
Change update mechanism of `geordi chromedriver-update`

niklas-hasselmeyer:
This command (and `geordi cucumber`/`geordi tests` if the `auto_update_chromedriver` option is active) will now always update to the latest version of chromedriver for the current chrome version.


denzelem (approved):
Thanks!

PR #175, opened 2022-05-25
use YAML.safe_load in dumple and allow aliases

dastra-mak:
In v9.1.0 in noticed that the `geordi dump` command fails in projects that include YAML aliases in the database.yml file:

```
g dump

# Dumping the development database
Unknown alias: test

x Something went wrong.
```
`database.yml`:
```
development:
  adapter: postgresql
  encoding: unicode
  host: 127.0.0.1
  database: foo_development

test: &test
  adapter: postgresql
  encoding: unicode
  host: 127.0.0.1
  database: foo_test<%= ENV['TEST_ENV_NUMBER']%>

cucumber:
  <<: *test
```

In my project Psych `v4.0.1` was used. The load method changed, to use `safe_load` instead.
As the psych version is not fixed in geordi, I added a version switch that can cope with new and old versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ## Unreleased
 
 ### Compatible changes
+* Use YAML.safe_load in dumple and allow aliases
 
 ### Breaking changes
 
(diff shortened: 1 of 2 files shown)


codener:
This change looks sane, thank you very much. One question though: Are you sure that `Psych` is available in each Ruby (2.5+) without any `require`?

dastra-mak:
Yes, it is part of stdlib from 1.9.2 on (see https://github.com/ruby/psych#installation)

codener (approved):

Issue #176, opened 2022-07-26
Disable multi-line autocomplete by default


vrandr:
Pasting in IRB 1.2+ is very slow. See these cards for further infos:

- https://makandracards.com/makandra/512584-irb-s-multi-line-autocomplete-and-how-to-disable-it
- https://makandracards.com/makandra/490022-how-to-fix-pasting-in-irb-1-2+-is-very-slow#section-option-2-disable-by-default

Can you disable the IRB multiline feature by calling `rails console -- --nomultiline` in https://github.com/makandra/geordi/blob/master/lib/geordi/util.rb#L86-L94?

This is supported. See https://github.com/rails/rails/issues/39909 for details.

Denionysos:
> Is there any decision on this issue, if this will be implemented in geordi?

Yes, we have decided to implement the proposal. The change will come in the next days

codener:
_Copied from PR:_

I tried geordi console on Ruby 2.7 with IRB 1.4.1 locally and IRB 1.2.6 on the Server. Unfortunately, pasting is still slow.

It seems the performance improvement is bound to the Ruby version and not the irb version. So please refactor this accordingly.

Please also update the CHANGELOG with notable changes (see https://github.com/makandra/geordi#contributing), and please print a message when using the flag:

```ruby
Interaction.note 'Using --no-multiline switch for faster pasting'
```

PR #182, opened 2022-11-23
disable multiline autocomplete by default #176

Denionysos:


features/console.feature
```diff
@@ -5,3 +5,16 @@ Feature: The console command
...
+      And a file named "config/deploy.rb" with "deploy file exists"
+      And a file named "config/deploy/staging.rb" with:
+    """
+    set :deploy_to, '/var/www/example.com'
+    set :user, 'user'
+    server 'www.example.com'
+    """
+    When I run `geordi console staging`
+    Then the output should contain "# Opening a Rails console on staging"
+      And the output should contain "Util.run! ssh, user@www.example.com, -t, cd /var/www/example.com/current && bash --login -c 'bundle exec rails console -e  -- --nomultiline"
```
codener, line 20: :ok_hand:

lib/geordi/util.rb
```diff
@@ -84,12 +84,14 @@ def binstub_or_fallback(executable)
       end
 
       def console_command(environment)
+        multiline_flag = irb_version >= Gem::Version.new('1.2') ? '-- --nomultiline' : ''
```
codener, line 87: Please also set an upper limit, as per [this card](https://makandracards.com/makandra/490022-how-to-fix-pasting-in-irb-1-2+-is-very-slow) the issue is fixed in 1.2.6.

Tip:
```
multiline_flag = if condition
  ...
end
```
will leave the variable with `nil` if the condition is `false`. It can be used right ahead in interpolation like `"console #{multiline_flag}"`.

lib/geordi/util.rb
```diff
@@ -84,12 +84,16 @@ def binstub_or_fallback(executable)
       end
 
       def console_command(environment)
+        multiline_flag = if irb_version >= Gem::Version.new('1.2') && irb_version < Gem::Version.new('1.2.6')
+                           '-- --nomultiline'
+                         end
+
         if gem_major_version('rails') == 2
-          'script/console ' + environment
+          "script/console #{environment}"
```
codener, line 92: :+1:

features/console.feature
```diff
@@ -1,7 +1,45 @@
 Feature: The console command
   Most aspects of connection to a server are tested in shell feature.
 
-  Scenario: Opening a local Rails console
+  Scenario: Opening a local Rails console with an irb version between 1.2 and 1.2.6
+    Given the irb version is "1.2"
```
codener, line 5: I don't think the Gem was ever released as "1.2". Please use "1.2.0" instead.

lib/geordi/util.rb
```diff
@@ -182,6 +186,16 @@ def testing?
         !!ENV['GEORDI_TESTING']
       end
 
+      def irb_version
+        version_string = if testing?
+                           ENV['GEORDI_TESTING_IRB_VERSION']
+                         else
+                           `irb --version`.match(/\d\.\d/)
```
codener, line 193: When you're only matching two digits, how can you distinguish "1.2" and "1.2.6"? This really is "good luck matching". Also, this does not return a string, but a MatchData object.

Better:
```suggestion
                           `irb --version`[/irb (\d\.\d\.\d)/, 1]
```
This is a shorthand for "match this regex against this string, and return what was matched by parentheses group no 1.

Please also reduce the indentation to two spaces.

features/console.feature
```diff
@@ -1,7 +1,45 @@
...
+    And a file named "config/deploy.rb" with "deploy file exists"
+    And a file named "config/deploy/staging.rb" with:
+    """
+    set :deploy_to, '/var/www/example.com'
+    set :user, 'user'
+    server 'www.example.com'
+    """
+    When I run `geordi console staging`
+    Then the output should contain "# Opening a Rails console on staging"
+    And the output should contain "Util.run! ssh, user@www.example.com, -t, cd /var/www/example.com/current && bash --login -c 'bundle exec rails console -e"
```
codener, line 45: This test would be green on v1.2.1 as well – you should add `But the output should not contain "nomultiline"` to all tests that expect the flag to be absent.

Note that negative expectations ("should not") are quite easy to satisfy. E.g. `the output should not contain "-- -nomultiline"` would pass for all three scenarios, by accident. Thus we aim to write them as general as possible.

features/console.feature
```diff
@@ -1,7 +1,45 @@
...
-  Scenario: Opening a local Rails console
+  Scenario: Opening a local Rails console with an irb version between 1.2 and 1.2.6
+    Given the irb version is "1.2"
     When I run `geordi console`
     Then the output should contain "# Opening a local Rails console"
-      And the output should contain "Util.run! bundle exec rails console -e development"
+      And the output should contain "Util.run! bundle exec rails console -e development -- --nomultiline"
+
+
+  Scenario: Opening a local Rails console with irb version >= 1.2.6
```
codener, line 11: You only need to do the version testing once. Since local is so much simpler, I would suggest to do it only here. However, please to a check for all three cases: below 1.2, between 1.2 and 1.2.6, and from 1.2.6.

features/console.feature
```diff
@@ -1,7 +1,45 @@
...
+    set :deploy_to, '/var/www/example.com'
+    set :user, 'user'
+    server 'www.example.com'
+    """
+    When I run `geordi console staging`
+    Then the output should contain "# Opening a Rails console on staging"
+    And the output should contain "Util.run! ssh, user@www.example.com, -t, cd /var/www/example.com/current && bash --login -c 'bundle exec rails console -e  -- --nomultiline"
+
+
+  Scenario: Opening a remote Rails console with irb version >= 1.2.6
```
codener, line 33: This scenario may be removed.

codener (changes requested):
I have some more remarks.

Please note that using _the local irb version_ to decide whether to use the flag _remotely_ might break in case the developer has a different Ruby version locally and tries to open a remote console. This should be rare though, and doing the check on the server would be hard, so let's just keep it like this.

features/console.feature
```diff
@@ -1,22 +1,30 @@
...
+    When I run `geordi console`
+    Then the output should contain "# Opening a local Rails console"
+      And the output should contain "Util.run! bundle exec rails console -e development -- --nomultiline"
 
   Scenario: Opening a local Rails console with irb version >= 1.2.6
     Given the irb version is "1.2.6"
     When I run `geordi console`
     Then the output should contain "# Opening a local Rails console"
     And the output should contain "Util.run! bundle exec rails console -e development"
+    But the output should not contain "-- --nomultiline"
```
codener, line 23: This is what I explained in the other comment: with negative expectations, do not be precise but as general as possible. Otherwise, they might easily become falsely satisfied.

```suggestion
    But the output should not contain "nomultiline"
```

features/console.feature
```diff
@@ -1,22 +1,30 @@
...
 
-  Scenario: Opening a local Rails console with an irb version between 1.2 and 1.2.6
-    Given the irb version is "1.2"
+  Scenario: Opening a local Rails console with an irb version < 1.2.0
+    Given the irb version is "1.1.0"
     When I run `geordi console`
     Then the output should contain "# Opening a local Rails console"
-      And the output should contain "Util.run! bundle exec rails console -e development -- --nomultiline"
+    And the output should contain "Util.run! bundle exec rails console -e development"
+    But the output should not contain "-- --nomultiline"
```
codener, line 9: Same here.

codener (approved):
One change is still missing, the rest is fine. You may merge after updating it.

codener (approved):

codener:
I tried `geordi console` on Ruby 2.7 with IRB 1.4.1 locally and IRB 1.2.6 on the Server. Unfortunately, pasting is still slow. 

It seems the performance improvement is bound to the Ruby version and not the irb version. So please refactor this accordingly.

Please also update the CHANGELOG with notable changes.

PR #186, opened 2022-12-19
fix ruby_version method in util.rb

Denionysos:


lib/geordi/util.rb
```diff
@@ -200,7 +200,7 @@ def irb_version
       end
 
       def ruby_version
-        testing? ? Gem::Version.new(ENV['GEORDI_TESTING_RUBY_VERSION']) : RUBY_VERSION
+        Gem::Version.new(testing? ? ENV['GEORDI_TESTING_RUBY_VERSION'] : RUBY_VERSION)
```
codener, line 203: To have 1 line per action, please split this into two lines: deriving the version, and wrapping it.

codener (changes requested):
Thanks for the update! Please remember to reference the issue in the commit message in order to connect it with this PR.

codener (approved):
Thanks, looks good!

PR #187, opened 2023-01-09
Ruby 3.2 compatibility

foobear:
`File.exists?` has been deprecated since Ruby 2.1 and removed in 3.2. It was only an alias for `File.exist?` anyway.


codener (approved):
LGTM :+1: 

I'd see this as a patch level update.

foobear:
Released as version 9.4.1

PR #188, opened 2023-01-19
Improve handling of tags for cucumber command

brunosedler:


lib/geordi/cucumber.rb
```diff
@@ -108,6 +111,15 @@ def command_line_options
         # last arg. Do it manually here.
         last_arg = argv.last
         args << last_arg if (last_arg && last_arg.start_with?('-'))
+
+        args.map do |option|
+          if option.match?(/\S\s\S/)
+            # put options with a whitespace in quotes. Relevant for negated tag options ('not @tag')
+            option.prepend('"').concat('"')
```
codener, line 118: Instead of duplicating the Aruba step, couldn't you wrap the Option with single quotes here?

features/cucumber.feature
```diff
@@ -147,6 +147,34 @@ Feature: The cucumber command
     But the output should not contain "other.feature"
 
 
+  Scenario: Running all cucumber features matching a given tag
+    Tags need to be passed differently than in vanilla cucumber where the syntax is --tags @tag.
+    With geordi you need to pass them like --tags=@tag to be parsed correctly.
```
codener, line 152: Instead of leaving this note to developers here, let's print a warning to the user if `cucumber` is invoked with `--tags` but without equals sign or without at sign.

Actually, doesn't cucumber.rb:103-104 enable `cucumber --tags @tag`, i.e. without equals sign?

lib/geordi/cucumber.rb
```diff
@@ -108,6 +111,15 @@ def command_line_options
         # last arg. Do it manually here.
         last_arg = argv.last
         args << last_arg if (last_arg && last_arg.start_with?('-'))
+
+        args.map do |option|
+          if option.match?(/\S\s\S/)
+            # put options with a whitespace in quotes. Relevant for negated tag options ('not @tag')
```
codener, line 117: What is `option` when we're calling `--tags='not @tag'`? Why do we need to wrap something in quotes?

Please rename `args` to `options`.

features/cucumber.feature
brunosedler: Well kind of, but that's only for args that are finally passed to the actual cucumber command.
In `lib/geordi/commands/cucumber.rb:55` we partition passed args according to whether they start with `-` or not. Args starting with `-` are considered command option, and those without are considered filepaths. If you don't pass tags with an `=` the tag itself will be considered a file.  

Then in `lib/geordi/cucumber.rb:13` we split the args at `=` and thus make use of the code you referenced.

lib/geordi/cucumber.rb
brunosedler: Yes I thought about that. I simply preferred double quotes in the output and did not want to make compromises because of testing. But if you like single quotes just as much or think it's not worth to duplicate the Aruba step for this, I'll change it.

features/cucumber.feature
brunosedler: But your right, it would help to print this text as warning

lib/geordi/cucumber.rb
codener: IMHO choosing ' or " is rather cosmetic, while duplication of a complex piece of code from a library is WET. If you don't mind, please change as suggested.

features/cucumber.feature
codener: Ah I see. Could make the `cucumber` command use the argument partitioning from `Geordi::Cucumber#command_line_options` et al.? This would really be in the spirit of your MR.

If you do, we only need to warn about `--tags` without an @. Please remove that Cucumber scenario comment then.

lib/geordi/cucumber.rb
codener: So we're 
1. separating arguments and options in the cucumber command file
2. merging them back into a flat array in `Geordi::Cucumber.run`
3. separating them again in `Geordi::Cucumber#command_line_options`
4. and merging them again in `Geordi::Cucumber#parallel_execution_command`??!

Could you please fix this?

lib/geordi/cucumber.rb
```diff
@@ -8,9 +8,17 @@
 
 module Geordi
   class Cucumber
+    def run(cucumber_options, options = {})
```
codener, line 11: These are not only options, but all arguments to the Cucumber command. Since we're in the Cucumber context, I'd leave out "cucumber_", and I'd generalize the name to "arguments":
```suggestion
    def run(arguments, options = {})
```

lib/geordi/cucumber.rb
```diff
@@ -8,9 +8,17 @@
...
+      cucumber_options.map! do |option|
+        if option.match?(/\S\s\S/)
+          # put options with a whitespace in quotes. Relevant for negated tag options ('not @tag')
+          "'#{option}'"
+        else
+          option
+        end
+      end
+      self.argv = cucumber_options
```
codener, line 20: This could be simplified to 
```suggestion
      self.argv = arguments.each do |arg|
        # Ensure arguments containing white space are kept together
        arg.replace %('#{arg}') if arg.match?(/\S\s\S/)
      end
```

lib/geordi/cucumber.rb
```diff
@@ -32,33 +40,43 @@ def serial_execution_command
...
-
       [
         'bundle exec parallel_test -t ' + type_arg,
-        %(-o '#{command_line_options.join(' ')}'),
+        %(-o "#{command_line_options.join(' ')}"),
         "-- #{features.join(' ')}",
       ].compact.join(' ')
     end
 
+    def filter_command_line_features!
```
codener, line 63:
```suggestion
    def drop_command_line_features!
```

lib/geordi/cucumber.rb
```diff
@@ -32,33 +40,43 @@ def serial_execution_command
...
     end
 
     def show_features_to_run
-      if command_line_options.include? 'rerun'
+      tags = command_line_options & ['--tags','-t']
+      if command_line_options.include?('rerun')
         Interaction.note 'Rerunning failed scenarios'
+      elsif command_line_options.include?('--tags') || command_line_options.include?('-t')
+        tag_index = command_line_options.find_index(tags.first) + 1
+        Interaction.note "Only features matching tag option #{command_line_options[tag_index]}"
```
codener, line 79: L74 should be moved above L78 IMHO. Alternatively, you could introduce a method `#command_line_tag_options` that even accounts for multiple tags options:
```
def command_line_tag_options
  [].tap do |tag_options|
    command_line_options.each_cons do |option, tags|
      tag_options << tags if option =~ /--tags|-t/
    end
  end
end
```

codener (commented):
This solution looks really clean. Congrats!

I have a few suggestions though.

lib/geordi/cucumber.rb
brunosedler: Unfortunately this does not work. `#each` simply returns the original array . But I can shorten the code to  
```suggestion
      self.argv = arguments.map do |arg|
        # Ensure arguments containing white space are kept together
        arg.match?(/\S\s\S/) ? %('#{arg}') : arg
      end
```

lib/geordi/cucumber.rb
brunosedler: I like your suggested method, good idea :+1:

PR #195, opened 2023-03-14
Add a owner filter for pivotal tracker stories

martinschaflitzl1:
When choosing a story for `geordi commit` or `geordi branch` it can get quite cumbersome if there are a lot of running stories for the current project. Most of the time, I am only interested  in the stories where I am the owner. This MR adds the possibility to configure a owner-filter for the story selection in the `~/.config/geordi/global.yml`.  
This functionality has been proven to be useful in my workflow, please consider adding it to geordi.


codener:
Thanks for the PR, it's a good idea. However, the global filter config feels to hidden. A `--mine` (`-m`) option would make obvious why the list is short, and the default case would remain unfiltered. WDYT?

martinschaflitzl1:
The default kinda still is a unfiltered list, because you have to configure the filter before it can filter stories.  
I very rarely check out a branch for / commit a story I do not own on the Tracker, so I´d rather have the default (with the config set) filtered. But I also don´t know if that is the case for all our colleges.   
A `-m` option would also be fine for me, I can always have an alias that filters by default.

codener:
Ok, let's do both then:
- `geordi [branch|commit]` shows all stories
- `geordi [branch|commit] --mine` (`-m`) shows only own stories

When the suggested global config is set:
- `geordi [branch|commit]` shows only own stories
- `geordi [branch|commit] --no-mine`  (`-M`) shows all stories

(The `--no-mine` is Thor default for boolean options.)

features/branch.feature
```diff
@@ -59,3 +61,19 @@ Feature: Check out a feature branch based on a story from Pivotal Tracker
...
+      And I type ""
+    Then the output should contain "Using the stories filter: 'state:started,finished,rejected owner:maxmusterman'"
+
+
+  Scenario: You can skip the owner filter
+    Given a file named "tmp/global_settings.yml" with "pivotal_tracker_api_key: my_api_key\npivotal_tracker_owner_filter: maxmusterman"
+
+    When I run `geordi branch --skip-owner-filter` interactively
+      And I type ""
+    Then the output should contain "Using the stories filter: 'state:started,finished,rejected'"
```
codener, line 79: Caution: you should add a step for `the output should not contain "owner:"`

features/branch.feature
```diff
@@ -10,6 +10,8 @@ Feature: Check out a feature branch based on a story from Pivotal Tracker
       # I skip the initials prompt
       And I type ""
     Then the output should contain "Util.run! git, checkout, -b, mm/test-story-12"
+      # if not configured, the stories are not filtered by owner
+      And the output should contain "Using the stories filter: 'state:started,finished,rejected'"
```
codener, line 14: Same here: rather replace this step with "should not contain owner:"

features/commit.feature
```diff
@@ -48,3 +50,19 @@ Feature: Creating a git commit from a Pivotal Tracker story
...
+      And I type ""
+    Then the output should contain "Using the stories filter: 'state:started,finished,rejected owner:maxmusterman'"
+
+
+  Scenario: You can skip the owner filter
+    Given a file named "tmp/global_settings.yml" with "pivotal_tracker_api_key: my_api_key\npivotal_tracker_owner_filter: maxmusterman"
+
+    When I run `geordi commit --skip-owner-filter` interactively
+      And I type ""
+    Then the output should contain "Using the stories filter: 'state:started,finished,rejected'"
```
codener, line 68: Same applies to this file.

lib/geordi/commands/commit.rb
```diff
@@ -6,9 +6,14 @@
 
 On the first execution we ask for your Pivotal Tracker API token. It will be
 stored in `~/.config/geordi/global.yml`.
+
+You can filter the stories by owner by adding `pivotal_tracker_owner_filter: <your pivotal tracker username>` to your `~/.config/geordi/global.yml`.
+This filter can be skipped with the `-o` option.
```
codener, line 11:
```suggestion
If you add `pivotal_tracker_username: $your_username` to `~/.config/geordi/global.yml`, this command will only consider stories owned by you.
```

lib/geordi/commands/commit.rb
```diff
@@ -6,9 +6,14 @@
...
 LONGDESC
 
+option :skip_owner_filter, aliases: '-o', type: :boolean, desc: 'Don´t filter stories by owner'
```
codener, line 14:
```suggestion
option :any_owner, aliases: '-a', type: :boolean, desc: 'Do not filter stories by owner'
```

lib/geordi/commands/branch.rb
```diff
@@ -4,11 +4,15 @@
...
+This filter can be skipped with the `-o` option.
 LONGDESC
 
 option :from_master, aliases: '-m', type: :boolean, desc: 'Branch from master instead of the current branch'
+option :skip_owner_filter, aliases: '-o', type: :boolean, desc: 'Don´t filter stories by owner'
 
 def branch
   require 'geordi/gitpt'
-  Gitpt.new.run_branch(from_master: options.from_master)
+  Gitpt.new(skip_owner_filter: options.skip_owner_filter).run_branch(from_master: options.from_master)
```
codener, line 17: Please update this command like commented on `commit`.

lib/geordi/gitpt.rb
```diff
@@ -76,15 +77,26 @@ def load_projects
 
     def applicable_stories
       if Util.testing?
+        puts "Using the stories filter: '#{stories_filter}'"
```
codener, line 80: Actually, it would be nice to notify the user about the owner filtering. Instead of putsing the filter here, could you always `note 'Filtering stories by owner "$owner"'` if stories are filtered?

codener (commented):
Looks good already. I have a few more remarks.

Also, please wait until #198 is merged (or rebase onto it), then add hints for the global config and for the skip option.

codener:
@martinschaflitzl1 Do you intend to finish this? Else we'll close it soon.

PR #198, opened 2023-07-07
[#183312781] Did you know? Geordi-Features in Geordi teasern

steevokovsky-mkd:


spec/did_you_know_spec.rb
```diff
@@ -0,0 +1,68 @@
+RSpec.describe Geordi::Hint do
+
+  subject { Geordi::Hint }
+
+  describe '#did_you_know' do
+
+    it 'prints a random hint in the console if an array with suitable suggestions was provided and returns all generated hints' do
```
codener, line 7: Since the return value is not relevant for the "Did you know" feature, it should not be mentioned in test descriptions. Also, the condition is not needed in the example description – it is a case that does not occur in the example. Suggested description:

```suggestion
    it 'generates hints from its arguments and prints one of them to stdout' do
```

spec/did_you_know_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      allow(Geordi::Hint).to receive(:hint).with(:branch).and_return('Did you know? `geordi branch` can check out a feature branch based on a story from Pivotal Tracker')
+      allow(Geordi::Hint).to receive(:hint).with([:branch, :from_master]).and_return('Did you know? `geordi branch -m` can branch from master instead of the current branch')
+      allow(Geordi::Hint).to receive(:hint).with('`geordi branch`').and_return('Did you know? `geordi branch`')
```
codener, line 10: You're stubbing the hint generation here. Why? The hints are the most important part of this method. I suggest to not stub them, but make them part of the test instead. In `describe '#hint'` below, you only need to care about edge cases then. (As I suggested above, #hint should probably be a private method, which we avoid to test.)

---
Generally on stubbing: you should avoid using real input and output when stubbing. The test _looks_ like normal, but all values were stubbed. When stubbing, always use made-up, self-explaining values like this:
```suggestion
      allow(Geordi::Hint).to receive(:hint).with(:command_1).and_return('Did you know 1')
      allow(Geordi::Hint).to receive(:hint).with(:command_2).and_return('Did you know 2')
```
Then in each part of your test, it gets immediately clear you're working with example data.

spec/did_you_know_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    it 'prints nothing to the console if an empty array was provided' do
+      array = []
+      expect {subject.did_you_know(array)}.not_to output(/.+/).to_stdout
+    end
+
+    it 'prints nothing to the console if an array with nil elements was provided' do
+      array = [nil, nil]
+      expect {subject.did_you_know(array)}.not_to output(/.+/).to_stdout
+    end
```
codener, line 33: Calling `Hint.did_you_know` with either no arguments or `nil` values should not silently proceed, but both crash in the face of the developer. It is not intended behavior, but misusage. In this light, you _might_ add a test that expects the exception, but I'd leave this out altogether.

codener (changes requested):
Thanks, looks really good already. I have a few editorial remarks, and a few notes about the tests. Please have a look!

features/rake.feature
```diff
@@ -8,3 +8,10 @@ Feature: The rake command
 
     When I run `geordi rake`
     Then the output should contain "Util.run! bin/rake"
+
+
+  Scenario: A hint is displayed after running the rake command
+    Given a file named "bin/rake" with "binstub"
+
+    When I run `geordi rake`
+    Then the output should contain "Did you know? `geordi capistrano` can run a capistrano command on all deploy targets"
```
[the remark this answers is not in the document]
steevokovsky-mkd: The description in capistrano.rb `desc 'capistrano COMMAND', 'Run a capistrano command on all deploy targets'` has Capistrano downcased, I'll change that too.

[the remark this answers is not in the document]
steevokovsky-mkd: I tried to stub `Geordi::Settings` to allow this feature to print hints (as we set the `settings_probability` to zero for tests):
```
Given(/^the settings allow hints to be printed$/) do
  allow_any_instance_of(Geordi::Settings).to receive(:did_you_know).and_return(100)
end
```

This does not work, if I debug this step, `did_you_know` actually returns 100, after that though it reverts to 0. I removed the feature as the spec already acts like a feature test (output and generated hints are tested).

lib/geordi/hint.rb
```diff
@@ -0,0 +1,46 @@
...
+        elsif hint.is_a?(Array)
+          command = Geordi::CLI.commands[hint[0].to_s]
+          option = command.options[hint[1]]
+          banner = option.banner.nil? ? '' : " #{option.banner}"
+          description = downcase_first_letter(option.description)
+          "Did you know? `geordi #{command.name} #{option.aliases.first}#{banner}` can #{description}"
+        elsif hint.is_a?(String)
+          "Did you know? #{hint}"
+        else
+          nil
```
codener, line 37: Please do not return `nil`, but `raise "Unsupported hint input #{hint.inspect}"` instead. This case is a true exception that should not occur. If it does, it should not be swallowed, but immediately crash instead.

features/rake.feature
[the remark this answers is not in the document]
codener: That's strange. Let's look at it next time you are here. I'd really like to have a Cucumber feature, as a unit test cannot really replace an integration test.

lib/geordi/commands/cucumber.rb
```diff
@@ -72,6 +72,13 @@ def cucumber(*args)
 
     Interaction.success 'Features green.'
 
+    Hint.did_you_know [
+      :rspec,
+      [:cucumber, :modified],
+      [:cucumber, :containing],
+      [:cucumber, :debug],
+      'Geordi can automatically update chromedriver before Cucumber tests. See geordi help chromedriver-update.'
```
codener, line 80: This should have backticks
```suggestion
      'Geordi can automatically update chromedriver before Cucumber tests. See `geordi help chromedriver-update`.'
```

lib/geordi/hint.rb
```diff
@@ -0,0 +1,46 @@
...
+
+module Geordi
+  class Hint
+    class << self
+
+      def did_you_know(hints)
+        settings_probability =  Settings.new.did_you_know
+        should_print_hint = Random.new.rand(100) <= (settings_probability || 10)
+
+        generated_hints = hints.map { |hint| generate(hint) }.compact
```
codener, line 12: This `.compact` is not needed any more. And _if you like_, you can simplify the code with:
```suggestion
        generated_hints = hints.map(&method(:generate))
```

codener (commented):
LGTM, I'm just missing a Cucumber feature.

features/did_you_know.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Some commands print hints about related commands and options
+
+  Scenario: A hint is displayed after running the rake command
+    Given a file named "bin/rake" with "binstub"
+    And a file named "tmp/global_settings.yml" with "did_you_know: 100"
```
codener, line 5: Now that I see this in practice, let's rename the config option once more: `hint_probability`.

lib/geordi/hint.rb
codener: Please remember to drop the `.compact`.

codener (commented):
Great. Please rename the config option as suggested, afterwards we should be ready to merge.

lib/geordi/hint.rb
```diff
@@ -0,0 +1,47 @@
+# Use the methods in this file to hint at other geordi features
+require File.expand_path('settings', __dir__)
+
+module Geordi
+  class Hint
+    class << self
+
+      def did_you_know(hints)
+        hint_probability =  Settings.new.hint_probability
+        default_probability = Util.testing? ? 0 : 10 # Percent
```
codener, line 10: Please keep this `settings_probability`, or, equally well:
```suggestion
        hint_probability =  Settings.new.hint_probability 
        hint_probability ||= (Util.testing? ? 0 : 10) # Default
```

codener (commented):
One last remark!

lib/geordi/hint.rb
```diff
@@ -0,0 +1,47 @@
+# Use the methods in this file to hint at other geordi features
+require File.expand_path('settings', __dir__)
+
+module Geordi
+  class Hint
+    class << self
+
+      def did_you_know(hints)
+        hint_probability =  Settings.new.hint_probability
+        default_probability ||= (Util.testing? ? 0 : 10) # Default
```
codener, line 10: Sorry for the iteration, but now you're `||=`-assigning an uninitialized variable. Please update once more:
```suggestion
        hint_probability = Settings.new.hint_probability
        default_probability = (Util.testing? ? 0 : 10)
```

codener: After clarification:
```suggestion
        settings_probability = Settings.new.hint_probability
        default_probability = (Util.testing? ? 0 : 10) # Percent
```

codener (approved):
Done! Please merge.

PR #199, opened 2023-07-07
Use local user for 'drop-databases' command

brunosedler:
These changes would make the `drop-databases` - command just access the postgresql and mysql shells as local user without any password validation. The  changes further assume, that the local user has the required permissions to drop the selected databases. These assumptions are correct for our current setup.

### DIFF
--- a/lib/geordi/commands/drop_databases.rb
+++ b/lib/geordi/commands/drop_databases.rb
@@ -7,6 +7,9 @@ and offer to delete them. Excluded are databases that are whitelisted. This come
 in handy when you're keeping your currently active projects in the whitelist files
 and perform regular housekeeping with Geordi.
 
+Per default, Geordi will try to connect to the databases as a local user without 
+password authorization.  
+
 Geordi will ask for confirmation before actually dropping databases and will
 offer to edit the whitelist instead.
 LONGDESC
@@ -19,6 +22,8 @@ option :postgres, banner: 'PORT_OR_SOCKET',
   desc: 'Use Postgres port or socket'
 option :mysql, banner: 'PORT_OR_SOCKET',
   desc: 'Use MySQL/MariaDB port or socket'
+option :sudo, aliases: '-S',  type: :boolean, default: false,
+ desc: 'Access databases as root'
 
 def drop_databases
   require 'geordi/db_cleaner'
@@ -42,9 +47,15 @@ def drop_databases
     postgres_flags = "--port=#{options.postgres}"
   end
 
-  extra_flags = { 'mysql' => mysql_flags,
-                 'postgres' => postgres_flags }
-  cleaner = DBCleaner.new(extra_flags)
+  unless options.sudo
+    Interaction.note 'Assuming your local user has permission to drop databases. Run with `--sudo` to use sudo.'
+  end
+
+  extra_flags = {
+    'mysql' => mysql_flags,
+    'postgres' => postgres_flags,
+  }
+  cleaner = DBCleaner.new(extra_flags, sudo: options.sudo)
   cleaner.clean_mysql unless options.postgres_only
   cleaner.clean_postgres unless options.mysql_only
 
--- a/lib/geordi/db_cleaner.rb
+++ b/lib/geordi/db_cleaner.rb
@@ -19,26 +19,26 @@ module Geordi
       @derivative_dbname = /_(test\d*|development|cucumber)$/
       base_directory = ENV['XDG_CONFIG_HOME']
       base_directory = Dir.home.to_s if base_directory.nil?
-      @whitelist_directory = File.join(base_directory, '.config', 'geordi', 'whitelists')
-      FileUtils.mkdir_p(@whitelist_directory) unless File.directory? @whitelist_directory
+      @allowlist_directory = File.join(base_directory, '.config', 'geordi', 'allowlists')
+      FileUtils.mkdir_p(@allowlist_directory) unless File.directory? @allowlist_directory
       @mysql_command = decide_mysql_command(extra_flags['mysql'])
       @postgres_command = decide_postgres_command(extra_flags['postgres'])
     end
 
-    def edit_whitelist(dbtype)
-      whitelist = whitelist_fname(dbtype)
-      whitelisted_dbs = if File.exist? whitelist
-        Geordi::Util.stripped_lines(File.read(whitelist))\
+    def edit_allowlist(dbtype)
+      allowlist = allowlist_fname(dbtype)
+      allowlisted_dbs = if File.exist? allowlist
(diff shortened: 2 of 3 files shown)


lib/geordi/db_cleaner.rb
```diff
@@ -103,40 +112,43 @@ def edit_whitelist(dbtype)
...
+          break if thread.value.exitstatus == 0
+          # sudo mysql was not successful, switching to mysql-internal user management
+          mysql_error = stderr.read.lines[0].chomp.strip.split[1]
+          if %w[1045 1698].include? mysql_error # authentication failed
+            cmd = 'mysql -uroot'
+            cmd << " #{extra_flags}" unless extra_flags.nil?
+            unless File.exist? File.join(Dir.home, '.my.cnf')
+              puts "Please enter your MySQL/MariaDB password for account 'root'."
+              Interaction.warn "You should create a ~/.my.cnf file instead, or you'll need to enter your MySQL root password for each db."
+              Interaction.warn 'See https://makandracards.com/makandra/50813-store-mysql-passwords-for-development for more information.'
```
codener, line 135: Please make this one a `note`.

lib/geordi/db_cleaner.rb
```diff
@@ -5,12 +5,21 @@
...
+      if @system_user
+        puts 'Please enter your sudo password if asked, for db operations as system users'
+        puts "We're going to run `sudo -u postgres psql` for PostgreSQL"
+        puts '               and `sudo mysql`            for MariaDB (which uses PAM auth)'
+        `sudo true`
+        Interaction.fail 'sudo access is required for database operations as database users' if $? != 0
+      else
+        puts "We're going to run `psql` for PostgreSQL"
+        puts '               and `mysql`            for MariaDB'
+        puts 'That means, we are assuming your local user has the required permissions to drop the selected databases'
```
codener, line 20: The "we're going to run" part is to assure the user he can entrust us with their sudo password. I'd leave that away in this case and just state the assumption. However, we cannot mention the command option here since we do not know which command will call us. We'll need to do this inside the command, so we can drop the `else` part here.

lib/geordi/commands/drop_databases.rb
```diff
@@ -42,9 +47,11 @@ def drop_databases
     postgres_flags = "--port=#{options.postgres}"
   end
 
-  extra_flags = { 'mysql' => mysql_flags,
-                 'postgres' => postgres_flags }
-  cleaner = DBCleaner.new(extra_flags)
+  extra_flags = {
```
codener, line 50: I'd print the assumption above this line:

```
Interaction.note 'Assuming standard makandra setup. Run with --no-default if this does not work for you.'
```

lib/geordi/commands/drop_databases.rb
```diff
@@ -19,6 +22,8 @@
   desc: 'Use Postgres port or socket'
 option :mysql, banner: 'PORT_OR_SOCKET',
   desc: 'Use MySQL/MariaDB port or socket'
+option :system_user, aliases: '-S',  type: :boolean, default: false,
+ desc: 'Connect to database as system user and request database passwords'
```
codener, line 26: Actually this option says "don't rely on standard makandra setup". What about calling it `--explicit`? 
```suggestion
option :explicit, aliases: '-e',  type: :boolean, default: false,
 desc: 'Access databases with sudo and ask for database password (MySQL)'
```
Or, we could go with a `default` option that's initially true:
```suggestion
option :default, aliases: '-d',  type: :boolean, default: true,
 desc: 'Rely on makandra standard setup'
```
It can be disabled by passing `--no-default`. That's even prettier in my eyes.

codener (commented):
LGTM. I have a few remarks.

lib/geordi/commands/drop_databases.rb
brunosedler: I think the `default` option and the description are to vague. I wouldn't be sure, what's encompassed in the default setup. Is maybe postgres default? I wouldn't know that we're talking about authorization for accessing the database.
This option is for people that don't use the standard setup, so imo it should tell explicitly what it does. If I use the default setup, the description doesn't offer any new value. If I don't use the default setup, the description is no help.

So I like the `explicit` suggestion better, but I'm not sure, if it really captures the meaning or implication of what the option is doing. What do you think about calling the option `login`?

```suggestion
option :login, aliases: '-l',  type: :boolean, default: false,
 desc: 'Access databases with sudo and ask for database password (MySQL)'
```

lib/geordi/commands/drop_databases.rb
brunosedler: I don't think it makes sense here. It would also be displayed, if you already ran the command with the `no-default` option. I think I'd leave it in the initializer in db_cleaner.rb, or at least put this in an if-block here

lib/geordi/db_cleaner.rb
```diff
@@ -103,40 +112,43 @@ def edit_whitelist(dbtype)
     end
 
     def decide_mysql_command(extra_flags)
-      cmd = 'sudo mysql'
+      cmd = @system_user ?'sudo mysql' : 'mysql'
```
codener, line 115: Missing a space here
```suggestion
      cmd = @system_user ? 'sudo mysql' : 'mysql'
```

lib/geordi/commands/drop_databases.rb
codener: We decided to go for `--sudo` with default `false`, but only if `--sudo` enables it and `--no-sudo` keeps it false.

codener: We decided to go for `--sudo` with default `false`, but only if `--sudo` enables it and `--no-sudo` keeps it false.

```suggestion
option :sudo, aliases: '-S',  type: :boolean, default: false,
 desc: 'Access databases as root'
```

lib/geordi/commands/drop_databases.rb
brunosedler: We decided to keep it here with a condition `if options.sudo`

codener:
Without the `--sudo` option, I'm getting 
> \# Checking for Postgres databases
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  database "codener" does not exist
> No postgres databases found that were not allowlisted.
 
Does this mean I have a nonstandard setup? Anyways, the error should be detected, i.e. the script should not continue with "no databases found". If possible, please suggest to use the --sudo option.

Also, renaming for "fair language" broke reference to previous allowlists/blocklists, which forces all users to create their lists again. As this also means the old lists with the "bad names" remain at their old locations, we should add migration code that renames the old lists.

Please also 
- Print a ">" before "Please enter your  ..." (#note method)
- Print "Nothing deleted." as #note as well (not #success)

PR #202, opened 2023-09-21
[Fix #200] update chromedriver upate routine

### LINKED ISSUE #200: Update chromedriver update routine
`ChromedriverUpdater` currently installs a chromedriver with the same version as Chrome. This strategy does not seem to hold any more, as [chromedriver announces](https://chromedriver.chromium.org/downloads):
> If you are using Chrome version 115 or newer, please consult[ the Chrome for Testing availability dashboard](https://googlechromelabs.github.io/chrome-for-testing/). This page provides convenient [JSON endpoints](https://googlechromelabs.github.io/chrome-for-testing/known-good-versions-with-downloads.json) for specific ChromeDriver version downloading.

Geordi will have to adopt the new way. However, we'll need to verify whether the new chromedriver-Chrome mapping will work with the normal Chrome, or require the [Chrome for Testing](https://googlechromelabs.github.io/chrome-for-testing/).

### Resources

- https://developer.chrome.com/blog/chrome-for-testing/
- https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints – chromedriver is included since v115.0.5763.0

tecden:
Tested locally that is actually downloads the latest version and unzip correctly.

### DIFF
--- a/lib/geordi/chromedriver_updater.rb
+++ b/lib/geordi/chromedriver_updater.rb
@@ -1,9 +1,12 @@
 require 'open3'
 require 'net/http'
 require 'tempfile'
+require 'json'
+require 'fileutils'
 
 module Geordi
   class ChromedriverUpdater
+    VERSIONS_PER_MILESTONES_URL = "https://googlechromelabs.github.io/chrome-for-testing/latest-versions-per-milestone-with-downloads.json"
 
     def run(options)
       chrome_version = determine_chrome_version
@@ -49,8 +52,8 @@ module Geordi
 
     # Check https://groups.google.com/a/chromium.org/g/chromium-discuss/c/4BB4jmsRyv8/m/TY3FXS4HBgAJ
     # for information how chrome version numbers work
-    def major_version(full_version)
-      full_version.match(/^(\d+\.\d+\.\d+)\.\d+$/)[1]
+    def milestone_version(full_version)
+      full_version.match(/^\d+/)[0]
     end
 
     def update_chromedriver(latest_chromedriver_version)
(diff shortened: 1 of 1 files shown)


lib/geordi/chromedriver_updater.rb
```diff
@@ -62,36 +65,65 @@ def update_chromedriver(latest_chromedriver_version)
...
+      latest_version || Interaction.fail("Could not parse chromedriver download data")
+    end
+
     def unzip(zip, output_dir)
       _stdout_str, _error_str, status = Open3.capture3('unzip', '-d', output_dir, '-o', zip.path)
 
+      # the archive contains a folder in which the relevant files are located. These files must be copied to ~/bin.
+      FileUtils.mv("#{output_dir}/chromedriver-linux64/chromedriver", output_dir)
+      FileUtils.mv("#{output_dir}/chromedriver-linux64//LICENSE.chromedriver", output_dir)
+      FileUtils.rm_rf("#{output_dir}/chromedriver-linux64")
```
codener, line 125: This should happen after the success check below. Also, the comment should say "must be moved", not copied.

lib/geordi/chromedriver_updater.rb
```diff
@@ -62,36 +65,65 @@ def update_chromedriver(latest_chromedriver_version)
...
+        rescue JSON::ParserError
+          Interaction.fail("Could not parse chromedriver download data")
+        end
+        @chromedriver_download_data = chromedriver_download_data
       end
     end
 
+    def latest_version(chrome_version)
+      latest_version = chromedriver_download_data.dig("milestones", milestone_version(chrome_version), "version")
+      latest_version || Interaction.fail("Could not parse chromedriver download data")
```
codener, line 116:
```suggestion
      latest_version || Interaction.fail("Could not find matching chromedriver version.")
```

codener (changes requested):
Looks really good! Just one edge case handling and two wordings annotated.

codener (approved):

PR #207, opened 2023-11-22
Run specs in github actions

brunosedler:


codener (approved):
Thanks for fixing this!

Issue #203, opened 2023-10-11
Disable multi-line autocomplete based on a user setting


makmic:
Hi :wave: 

I noticed that the bug fix for https://github.com/makandra/geordi/issues/176 does not in fact disable multi-line autocompletion by default but only for a specific version constraint:

```rb
irb_version >= Gem::Version.new('1.2') && ruby_version < Gem::Version.new('3.0')
```

Personally I really dislike the autocompletion feature and would like geordi to never enable it for the `geordi console <environment>`  command. To get there, I'd either have to modify the `~/.irbrc` of every deployment and set `IRB.conf[:USE_MULTILINE] = false` (not feasible and probably not compatible with other dev's preferences).. or ask you to add a geordi setting :-)

### Feature request
* Geordi exposes a new (project wide or global) setting for the `.geordi.yml` file
* E.g. 
```yml
use_multiline: false
```
* If the setting is `false`, `geordi console` should always use the `-- --nomultiline` flag

What do you think?

PR #209, opened 2023-11-29
[#203] add global config for irb flags

tecden:


tecden:
I tested with a Rails 7 application that the --nomultiline flag from the irb config is also correctly passed on to the console in the staging instance :heavy_check_mark:

lib/geordi/util.rb
```diff
@@ -89,14 +89,19 @@ def console_command(environment)
...
-          "#{binstub_or_fallback('rails')} console -e #{environment} #{use_multiline_flag}"
+          "#{binstub_or_fallback('rails')} console -e #{environment} #{irb_flags}"
         end
       end
 
+      def irb_flags
+        irb_flags = [use_multiline_flag, Settings.new.irb_flags].join(' ').strip
+        "-- #{irb_flags}" unless irb_flags.empty?
+      end
+
```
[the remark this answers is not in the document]
tecden: Personally, I think that the code is much more readable if the logic for the IRB flags sits in its own method. 
What do you think about putting the methods `#irb_flag` and `#use_multiline_flag` in the private scope of the Util class?

lib/geordi/commands/console.rb
```diff
@@ -6,6 +6,8 @@
 
 Selecting the server: `geordi console staging -s` shows a menu with all available
 servers. When passed a number, directly connects to the selected server.
+Irb flags defined in the global Geordi config file (`~/.config/geordi/global.yml`),
+are automatically passed to the irb console.
```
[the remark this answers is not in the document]
tecden: do you mean Geordi::Settings::GLOBAL_SETTINGS_FILE_NAME ?
I don't think it's a good idea to write this in the long description. The constant is prefixed with ENV['HOME']. And then my path is written into the global readme?

Or did you mean a different constant?

lib/geordi/util.rb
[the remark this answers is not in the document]
codener: As discussed, please implement as suggested.

lib/geordi/commands/console.rb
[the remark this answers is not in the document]
codener: Oh, that's a good point. In this case, just leave it like this.

lib/geordi/util.rb
```diff
@@ -94,14 +94,15 @@ def console_command(environment)
...
         else
-          "#{binstub_or_fallback('rails')} console -e #{environment} #{use_multiline_flag}"
-        end
-      end
+          use_multiline = if irb_version >= Gem::Version.new('1.2') && ruby_version < Gem::Version.new('3.0')
+            Interaction.note 'Using --nomultiline switch for faster pasting'
+            '--nomultiline'
+          end
+
+          irb_flags = [use_multiline, Settings.new.irb_flags].join('').strip
```
codener, line 102: Wouldn't you need to join with a `' '` space?! There should be a test covering this.

features/console.feature
```diff
@@ -1,6 +1,21 @@
...
+      And a file named "config/deploy.rb" with "deploy file exists"
+      And a file named "config/deploy/staging.rb" with:
+      """
+      set :deploy_to, '/var/www/example.com'
+      set :user, 'user'
+      server 'www.example.com'
+      """
+    When I run `geordi console staging`
+    Then the output should contain "# Opening a Rails console on staging"
+      And the output should contain "Util.run! ssh, user@www.example.com, -t, cd /var/www/example.com/current && bash --login -c 'bundle exec rails console -e  -- --foo --baz"
```
codener, line 16: There is the environment missing after `-e` (shouldn't it be "staging"?), and/or a double space present. Please show how `--nomultiline` is integrated.

lib/geordi/commands/console.rb
```diff
@@ -6,6 +6,9 @@
 
 Selecting the server: `geordi console staging -s` shows a menu with all available
 servers. When passed a number, directly connects to the selected server.
+
+IRB flags defined in the global Geordi config file (`~/.config/geordi/global.yml`),
+are automatically passed to the irb console.
```
codener, line 11: We must show how to use the setting:
```suggestion
IRB flags can be given as `irb_flags: '...'` in the global Geordi config file
(`~/.config/geordi/global.yml`).
```

features/console.feature
tecden: You are right. The environment is pulled out of the capistrano config. In the test setup the `rails_env` is missing in `config/deploy/staing.rb`. I have adjusted the test setup accordingly.

lib/geordi/commands/console.rb
```diff
@@ -7,8 +7,7 @@
 Selecting the server: `geordi console staging -s` shows a menu with all available
 servers. When passed a number, directly connects to the selected server.
 
-IRB flags defined in the global Geordi config file (`~/.config/geordi/global.yml`),
-are automatically passed to the irb console.
+IRB flags can be given as `irb_flags: '...' in the global Geordi config file (`~/.config/geordi/global.yml`).
```
codener, line 10:
```suggestion
IRB flags can be given as `irb_flags: '...'` in the global Geordi config file (`~/.config/geordi/global.yml`).
```
Tip: review README changes (needs update!) in the Markdown preview.

codener (approved):

PR #210, opened 2023-12-01
Improve/fix interrupt handling in locale Rails console

codener:
Previously, hitting Ctrl + C in a Rails console started by `geordi console`, the shell would get stuck between bash and console, effectively broken. By using `exec` instead of `system`, this can be fixed without having to implement SIGINT handling in Geordi.

See story #186280123.

### DIFF
--- a/lib/geordi/util.rb
+++ b/lib/geordi/util.rb
@@ -37,7 +37,8 @@ module Geordi
       # show_cmd: Whether to print the command
       # confirm: Whether to ask for confirmation before running it
       # fail_message: The text to print on command failure
-      def run!(command, show_cmd: false, confirm: false, fail_message: 'Something went wrong.')
+      # exec: Whether to run the command with `exec` instead of `system`
+      def run!(command, show_cmd: false, confirm: false, fail_message: 'Something went wrong.', exec: false)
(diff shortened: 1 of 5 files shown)


lib/geordi/util.rb
```diff
@@ -60,19 +61,23 @@ def run!(command, show_cmd: false, confirm: false, fail_message: 'Something went
...
-              system(*command)
+              Kernel.public_send(method, *command)
             end
           else
-            Bundler.clean_system(*command)
+            method = exec ? :clean_exec : :clean_system
+            Bundler.public_send(method, *command)
           end
 
+          # This part will never be reached when `exec` is true
```
tecden, line 80: Had to google once to find out how `exec` works. 

For my understanding: This part is never reached because Kernel#exec replaces the current process (geordi) with the process called by exec (rails console)?

tecden (approved):
lgtm

lib/geordi/util.rb
codener: That's right, see https://makandracards.com/makandra/1243-execution-of-shell-code-in-ruby-scripts#section-exec-show-archive-org-snapshot-the-last-thing-you-do

PR #213, opened 2024-02-14
Use dbdrop and dbcreate before restoring a postgres dump

brunosedler:


lib/geordi/dump_loader.rb
```diff
@@ -62,12 +62,16 @@ def mysql_command
...
     def postgresql_command
       ENV['PGPASSWORD'] = config['password']
-      command = 'pg_restore --no-owner --clean --no-acl'
-      command << ' --username=' << config['username'].to_s if config['username']
-      command << ' --port=' << config['port'].to_s if config['port']
-      command << ' --host=' << config['host'].to_s if config['host']
-      command << ' --dbname=' << config['database'].to_s
-      command << ' ' << dump_file
+      drop_command = 'dropdb --if-exists ' << config['database'].to_s
+      create_command = 'createdb ' << config['database'].to_s
```
codener, line 66: TIL!

However, `dropdb` and `createdb` have only been added in Postgres 12. This change will break restoring Postgres dumps for all folks on Postgres 11 and below. However, [Postgres 11 apparently is EOL](https://www.percona.com/blog/postgresql-11-will-soon-reach-end-of-life/), so probably this is fine.

codener (approved):
LGTM. Have you tried it locally?

brunosedler:
> LGTM. Have you tried it locally?

Yes, but not in a docker setup.

lib/geordi/dump_loader.rb
brunosedler: dropdb is shipped with postgresql-client package, so any postgres user should have it.

Issue #216, opened 2024-05-13
Improve flexibilty of `geordi tests`


makmic:
The current implementation of `geordi tests` looks like this:

```rb
def tests(*args)
  if args.any?
    args, opts = Thor::Options.split(args)
    error_message = "When passing arguments, the first argument must be either an RSpec or a Cucumber path."

    if args.empty?
      Interaction.fail error_message
    elsif args.first.start_with? 'spec'
      invoke 'rspec', args, opts
    elsif args.first.start_with? 'features'
      invoke 'cucumber', args, opts
    else
      Interaction.fail error_message
    end
  # ...
end
```

I wanted to run a command like this, which is not currently possible: 
> geordi test ./spec/config/i18n_spec.rb features/translation.feature

* The code above does not recognise rspec files within `./spec` (see `.start_with?('spec')`)
* The command does not allow specs and features to be mixed

As we're using GitLab which offers a quite practical [copy failed tests](https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/assets/javascripts/vue_merge_request_widget/extensions/test_report/index.vue#L97) button, I'd really be looking forward to support its output with the `geordi tests` command.

A possible implementation could look like this (untested!):

```rb
def tests(*args)
  if args.any?
    args, opts = Thor::Options.split(args)
    rspec_paths = []
    cucumber_paths = []

    args.each do |spec_path|
      if spec_path.end_with?('.feature')
        cucumber_paths << spec_path
      elsif spec_path.end_with?('_spec.rb')
        rspec_paths << spec_path
      else
        return Interaction.fail "When passing arguments, each argument must be either be a RSpec or a Cucumber path."
      end
    end

    if opts.any? && rspec_paths.any? && cucumber_paths.any?
      Interaction.fail "Cannot mix specs and features while passing RSpec or Cucumber CLI flags - we're unable to guess which CLI option belongs to which test runner."
    else
      invoke 'rspec', rspec_paths, opts if rspec_paths.any?
      invoke 'cucumber', cucumber_paths, opts if cucumber_paths.any?
    end
  # ...
end
```

WDYT?

codener:
Talked to @makmic. He just wants to be able to run all failed CI-tests locally with minimum effort.

Since we already have this command in place, adding this seems reasonable to me. Implementation notes:
[...]

PR #220, opened 2024-06-14
Improve flexibility of geordi test

brunosedler:
Closes https://github.com/makandra/geordi/issues/216


codener (approved):
> I assume if rspec fails, cucumber would not even be executed. That's the current behavior

Ok, right. Fine then!

PR #222, opened 2024-07-30
Add support for parallel_tests to "geordi unit"

martinschaflitzl1:
This will check and use parallel_tests for `geordi unit`. 

I also upgraded nokogiri, as the old version used would not install on my Ubuntu 22.04 installation.

### DIFF
--- a/features/unit.feature
+++ b/features/unit.feature
@@ -1,0 +1,19 @@
+Feature: The unit command
+
+  Background:
+    Given a file named "test/test_helper.rb" with "enable unit tests"
+
+  Scenario: A rake binstub is used if present
+    Given a file named "bin/rake" with "binstub"
+
+    When I run `geordi unit`
+    Then the output should contain "Util.run! bin/rake, test"
+
+
+  Scenario: A rake binstub is used to run parallel tests if present
+    Given a file named "bin/rake" with "binstub"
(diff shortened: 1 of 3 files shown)


features/unit.feature
```diff
@@ -0,0 +1,19 @@
+Feature: The rspec command
```
codener, line 1: Not rspec, or?

lib/geordi/commands/unit.rb
```diff
@@ -1,11 +1,23 @@
 desc 'unit', 'Run Test::Unit'
+long_desc <<-LONGDESC
+Runs Test::Unit with parallel_tests detection and `bundle exec`.
+
+In order to limit processes in a parallel run, you can set an environment
+variable like this: `PARALLEL_TEST_PROCESSORS=6 geordi unit`
```
codener, line 6: The LONGDESC will be rendered right below the short description in the README. Currently, it is a little redundant. What about:

```suggestion
Supports `parallel_tests`, binstubs and `bundle exec`.

In order to limit processes in a parallel run, you can set an environment
variable like this: `PARALLEL_TEST_PROCESSORS=6 geordi unit`
```

codener (approved):
I cannot judge whether this change will work in all usage scenarios. I don't use `geordi unit` myself. But by itself, it looks all good.

Please process the remarks, then merge.

Issue #219, opened 2024-05-29
Add support for Linear.app for `branch`-command


brunosedler:
In the future we will be using [linear.app](https://linear.app/) to manage stories and issues. Therefore geordi branch should be able to use linear issue titles to generate a new branch.

There is a ruby cli tool for the linear API: https://github.com/rubyists/linear-cli

codener:
See https://github.com/wulffeld/linear-branch-creator for inspiration.

Issue #218, opened 2024-05-29
Add support for Linear.app for `commit`-command


brunosedler:
In the future we will be using [linear.app](https://linear.app/) to manage stories and issues. Therefore `geordi commit` should be able to use linear issue titles to generate commit messages.

There is a ruby cli tool for the linear API: https://github.com/rubyists/linear-cli

makmic:
Minimal implementation suggestion (no dependencies)

Based on the graphql explorer https://studio.apollographql.com/public/Linear-API/variant/current/explorer 

```rb
def my_issue_titles(linear_team_key)
  filter = {
    "team": {
      "key": {
        "eq": linear_team_key,
      }
    },
    "assignee": {
      "isMe": {
        "eq": true
      }
    },
    "state": {
      "type": {
        "eq": "started"
      }
    }
  }
  response = request_with_payload(<<~GRAPHQL, filter:)
    query Issues($filter: IssueFilter) {
      issues(filter: $filter) {
        nodes {
          title
        }
      }
    }
  GRAPHQL

  response.dig(*%w[data issues nodes])&.map {|node| node['title'] }
end

API_ENDPOINT = 'https://api.linear.app/graphql'.freeze

def request_with_payload(attributes, variables)
  def client
  response = HTTP
   .headers({
      'Content-Type' => 'application/json',
      # API keys can be generated here: https://linear.app/makandra/settings/api
      'Authorization' => settings.linear_api_key
    })
   .post(API_ENDPOINT, json: { query: attributes.split.join(' '), variables: })
  @last_response = response
  parsed_response = JSON.parse(response.body)
  if parsed_response.key?('errors')
    raise parsed_response.dig('errors')
  else
    parsed_response['data']
  end
end
```

(I didn't test this particular code, ping me if you need further help)

makmic:
Hi :wave: ,

could you give me some rough estimate or info on the implementation timeline for this (and its sibling) issue?

Thx!

codener:
Closed by https://github.com/LisaHader/geordi/commit/13a4df48071e4a91fdf6f4dffd05fbb9fc8cba76.

codener:
Hi @makmic – sorry, didn't see your question. This feature should be released this week.

PR #226, opened 2024-10-07
[#218] Add support for Linear.app for commit-command; [#219] Add support for Linear.app for branch-command

LisaHader:


lib/geordi/commands/branch.rb
```diff
@@ -1,16 +1,16 @@
-desc 'branch', 'Check out a feature branch based on a story from Pivotal Tracker'
+desc 'branch', 'Check out a feature branch based on an issue from Linear'
```
codener, line 1:
```suggestion
desc 'branch', 'Check out a feature branch based on a Linear issue'
```

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
+require 'highline'
+require 'net/http'
+require 'json'
+require 'active_support/core_ext/object/blank'
```
codener, line 4: I think we cannot rely on active_support being around. Can you make it without?

features/step_definitions/miscellaneous_steps.rb
```diff
@@ -14,23 +14,18 @@
...
-  ENV['GEORDI_TESTING_GIT_USERNAME'] = username
-end
-
 Given /^my local git branches are: (.*)$/ do |branches|
   ENV['GEORDI_TESTING_GIT_BRANCHES'] = branches.split(", ").join("\n") + "\n"
 end
 
-Given 'there are no stories' do
-  ENV['GEORDI_TESTING_NO_PT_STORIES'] = 'true'
+Given 'there are no issues' do
```
codener, line 21:
```suggestion
Given 'there are no Linear issues' do
```

features/branch.feature
```diff
@@ -1,61 +1,33 @@
...
 
 
   Scenario: Checkout an existing branch with geordi branch
-    Given my username from git config is "Max Musterman"
-      And my local git branches are: master, mm/test-story-12
+    Given my local git branches are: master, mm/test-issue-12
 
     When I run `geordi branch` interactively
-      # I skip the initials prompt
       And I type ""
```
codener, line 26: The comment explains L42. If you remove it, I assume L42 can be removed as well. Maybe the command won't need to be run "interactively" any more? Would simplify the test under the hood.

Please check if this applies to other places as well.

tecden (commented):
I have two comments and small suggestions.
Tested geordi branch & geordi commit once. Worked 👍
Please add an entry to the changelog.
Thank you!

features/branch.feature
```diff
@@ -1,61 +1,33 @@
...
     Then the output should not contain "Util.run! git, checkout, master"
-      And the output should contain "Util.run! git, checkout, mm/test-story-12"
+      And the output should contain "Util.run! git, checkout, mm/test-issue-12"
```
codener, line 28: I am afraid this test is wrong now, and that is in part due to the weak scenario title. How can the test be green right now, if the branch name does not match the new Linear branch name pattern? Please investigate.

If I got it right from the current code, the scenario should be named "Scenario: If the target branch already exists, do not attempt to create a new one but simply switch to the existing branch". So please:
- Rename the branch to the new branch naming pattern
- Remove L27. While it is correct, it is also misleading.

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+    require File.expand_path('settings', __dir__)
+
+    API_ENDPOINT = 'https://api.linear.app/graphql'.freeze
+
+    def initialize
+      self.highline = HighLine.new
+      self.settings = Settings.new
+    end
+
+    def run_commit(git_args)
```
codener, line 19: Please remove the `run_` prefixes.

codener, line 19: Please remove the `run_` prefixes.

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+      Interaction.warn <<~WARNING unless Util.staged_changes?
+        No staged changes. Will create an empty commit.
+      WARNING
+
+      issue = choose_issue
```
codener, line 24: Maybe we can fail here just like in L31? Would give more feedback to the user. And renders the `if` obsolete :)

codener: If you agree, we should move the `fail` into `choose_issue`. WDYT?

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+      if issue
+        create_commit "[#{issue['identifier']}] #{issue['title']}", "Issue: #{issue['url']}", *git_args
+      end
+    end
+
+    def run_branch(from_master: false)
+      issue = choose_issue || Interaction.fail('No issue selected.')
+
+      branch_list_string = if Util.testing?
+                             ENV['GEORDI_TESTING_GIT_BRANCHES'] || ''
```
codener, line 34:
```suggestion
                             ENV['GEORDI_TESTING_GIT_BRANCHES'].to_s
```
should be just as fine.

lib/geordi/gitlinear.rb
codener: I think you don't even use it :)

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+                           else
+                             `git branch --format="%(refname:short)"`
+                           end
+
+      if branch_list_string.nil? || branch_list_string.strip.empty?
+        Interaction.fail 'Could not determine local git branches.'
```
codener, line 40:
```suggestion
        Interaction.fail 'Could not determine local Git branches.'
```

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+      end
+
+      new_branch_name = "#{issue['branchName']}"
+
+      local_branches = branch_list_string.split("\n")
```
codener, line 45: `local_branch_names` should become its own method, returning a list of all local branches, and failing if it cannot get them. It should memoize them like this:
```ruby
def local_branch_names
  @local_branch_names ||= begin
    # Fetch them
  end
end
```

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+                             ENV['GEORDI_TESTING_GIT_BRANCHES'] || ''
+                           else
+                             `git branch --format="%(refname:short)"`
+                           end
+
+      if branch_list_string.nil? || branch_list_string.strip.empty?
+        Interaction.fail 'Could not determine local git branches.'
+      end
+
+      new_branch_name = "#{issue['branchName']}"
```
codener, line 43: Why not like this? We could even let go of this variable, `issue['branchName']` is perfect.
```suggestion
      new_branch_name = issue['branchName']
```

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+
+      branch_name = local_branches.find { |branch_name| branch_name == new_branch_name }
+      branch_name ||= local_branches.find { |branch_name| branch_name.include? issue['identifier'].to_s }
+
+      if branch_name.present?
+        checkout_branch branch_name, new_branch: false
+      else
+        checkout_branch new_branch_name, new_branch: true, from_master: from_master
+      end
+    end
```
codener, line 55: This `if` block has the same decision that is made again inside `checkout_branch`. And that method is only used here. Please inline it.

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+
+      if branch_list_string.nil? || branch_list_string.strip.empty?
+        Interaction.fail 'Could not determine local git branches.'
+      end
+
+      new_branch_name = "#{issue['branchName']}"
+
+      local_branches = branch_list_string.split("\n")
+
+      branch_name = local_branches.find { |branch_name| branch_name == new_branch_name }
```
codener, line 47: The variable should be called after what it is: `matching_local_branch`. The whole method becomes a lot more readable by this.

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+      loading_message = 'Connecting to Linear ...'
+      print(loading_message)
+      issues = applicable_issues
+      reset_loading_message = "\r#{' ' * (loading_message.length + issues.length)}\r"
+
+      Geordi::Interaction.fail('No issues to offer.') if issues.empty?
+
+      if Util.testing?
+        return issues[0]
+      end
```
codener, line 71: If tests exit here anyway, there's no much gain in having a test switch in `applicable_issues` as well. Instead, please move the `Util.testing?` to the top of this method and do all required stuff directly there:
- Invoke `settings.linear_api_key` (presumably to see if settings are present)
- Return what L116+ does
Also, should the test identifier not be "team-123" for Linear? Afaik they are not numbers any more.

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+      nil # Return nothing
+    end
+
+    def create_commit(title, description, *git_args)
+      extra = highline.ask("\nAdd an optional message").strip
+      title << ' - ' << extra if extra != ''
+      Util.run!(['git', 'commit', '--allow-empty', '-m', title, '-m', description, *git_args])
+    end
+
+    def applicable_issues
```
codener, line 113: Please name this method so it is clear what it does:
```suggestion
    def fetch_linear_issues
```

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+
+      request = Net::HTTP::Post.new(uri.path)
+      request.body = "[ #{query} ]"
+
+      request['Content-Type'] = 'application/json'
+      request['Authorization'] = settings.linear_api_key
+
+      response = https.request(request)
+
+      @last_response = response
```
codener, line 177: What is this for?

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+            }
+          }
+        }
+      GRAPHQL
+
+      response.dig(*%w[issues nodes])
+
+    end
+
+    def request_with_payload(attributes, variables)
```
codener, line 161: Please be more specific about what this method does:
```suggestion
    def query_api(attributes, variables)
```

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+        },
+        "state": {
+          "type": {
+            "eq": "started"
+          }
+        }
+      }
+      response = request_with_payload(<<~GRAPHQL, filter: filter)
+        query Issues($filter: IssueFilter) {
+          issues(filter: $filter) {
```
codener, line 139: I have no experience in using GraphQL. Are all these "filters" necessary?

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+    def request_with_payload(attributes, variables)
+      uri = URI(API_ENDPOINT)
+
+      https = Net::HTTP.new(uri.host, uri.port)
+      https.use_ssl = true
+
+      query = { query: attributes.split.join(' '), variables: variables}.to_json
+
+      request = Net::HTTP::Post.new(uri.path)
+      request.body = "[ #{query} ]"
```
codener, line 170: Wouldn't you wrap the brackets around L167?
```ruby
      query = [{ query: attributes.split.join(' '), variables: variables}].to_json
```
Then here
```suggestion
      request.body = query
```

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+      response = https.request(request)
+
+      @last_response = response
+
+      parsed_response = JSON.parse(response.body)[0]
+      if parsed_response.key?('errors')
+        raise parsed_response.dig('errors')
+      else
+        parsed_response['data']
+      end
```
codener, line 184: Great :ok_hand:

lib/geordi/gitlinear.rb
```diff
@@ -0,0 +1,201 @@
...
+
+    attr_accessor :highline, :settings
+
+    def choose_issue
+      loading_message = 'Connecting to Linear ...'
+      print(loading_message)
+      issues = applicable_issues
+      reset_loading_message = "\r#{' ' * (loading_message.length + issues.length)}\r"
+
+      Geordi::Interaction.fail('No issues to offer.') if issues.empty?
```
codener, line 67: Wouldn't Geordi need to first `print reset_loading_message` before executing the `fail`?

codener (changes requested):
Hi Lisa,
thanks for implementing this. It looks good!

I have collected quite many comments. Most address code that you haven't written yourself, I think. Many are about code style with maintainability in mind. I found just a few important items, please have a look. If in doubt, just ask me.

Best regards

features/branch.feature
codener: Ah, the test is green because of some fallback logic. However, this test is not about the fallback logic, so it should not occur here.

lib/geordi/gitlinear.rb
LisaHader: Thanks for the suggestion! I moved the fail to `choose_issue`, that actually looks much cleaner.

lib/geordi/gitlinear.rb
LisaHader: You are right, I renamed the identifier to "team-123"

lib/geordi/gitlinear.rb
LisaHader: I added the `print` statement before the fail. Thanks for pointing that out!

lib/geordi/gitlinear.rb
LisaHader: `require 'active_support/core_ext/object/blank'` was needed for `if branch_name.present?` in L49, as `.present?` is defined in it. I replaced `.present?` with `.nil?`. IMO this should also be sufficient since `.find` returns `nil` by default if no object matches. Can you please take a quick look at it to make sure that's also okay here?

lib/geordi/settings.rb
```diff
@@ -132,54 +122,24 @@ def save_global_settings
...
-    def array_wrap_project_ids(project_ids)
-      case project_ids
+    def array_wrap_team_ids(team_ids)
+      case team_ids
       when Array
-        project_ids
+        team_ids
       when String
-        project_ids.split(/[\s]+/).map(&:to_i)
+        team_ids.split(/[\s]+/).map(&:to_i)
```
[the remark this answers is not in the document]
LisaHader: Could you please have another look at my solution? I normally would not test a private method.

My other approach would be testing the method Geordi::Settings.linear_team_ids but I was not sure on how to force @local_settings['linear_team_ids'] and @global_settings['linear_team_ids'] to contain the entries I want to test on.

lib/geordi/gitlinear.rb
LisaHader: I will have a look at this at Monday and come back to you about it. I do not have time to research this at this moment. Maybe there is a shorter solution

LisaHader: I did a bit of research on the topic by now and I do not think there is any better way to do this. This is the recommended way for inserting variabes by [GraphQL](https://graphql.org/learn/queries/#variables).

Alternatively, you could insert the filter directly into the query like it is described in the [Linear API](https://developers.linear.app/docs/graphql/working-with-the-graphql-api/filtering). But since our filter is a bit longer and more complex than the examples, I feel like the query gets more readable if we keep the filter as a seperate variable.

lib/geordi/gitlinear.rb
codener: I think you could write `branch_list_string.to_s.strip.empty?` to cover all cases (no `.nil?` check needed).

lib/geordi/gitlinear.rb
codener: Ok, fine then.

codener (approved):
LGTM!

codener (changes requested):
Code is fine now! However, I tried it locally (see CONTRIBUTING) and have a few more remarks:
- `note` must be `Interaction.note`
- Let's try and remove that loading message. It screws output and might not be needed any more (Pivotal Tracker API used to be very slow). Should the Linear API prove slow as well, we might add it back later. Bad output example:
  ```
  Connecting to Linear ...> No team id found.
  > Please open a team in Linear
  ```
- When asking for the API key, please use this prompt: `Please enter the API key now: `. After storing it, please print `note 'API key stored in /home/dominik.schoeler/.config/geordi/global.yml .'`.

codener:
Two more remarks after having a closer look at the issue menu:
- Please use "unassigned" instead of "none" when there is no assignee
- Please do not color the issue state for now. We might re-color later.

codener (approved):
Done! Thank you very much.

PR #231, opened 2025-01-27
Document how to change the CLI editor used by geordi

makmic:
I was surprised that `geordi drop-databases` uses `nano`, while my `$EDITOR` is set to `vim`.

Looking at geordis source code did not help me at first, but I found the solution to my problem in a PR description. With this patch it should be slightly easier for the next person. The `raise` branch is unlikely to be ever evaluated on a standard unix machine.

### DIFF
--- a/lib/geordi/util.rb
+++ b/lib/geordi/util.rb
@@ -137,14 +137,15 @@ module Geordi
         end
       end
 
-      # try to guess user's favorite cli text editor
       def decide_texteditor
-        %w[/usr/bin/editor vi].each do |texteditor|
-          if cmd_exists?(texteditor) && texteditor.start_with?('$')
-            return ENV[texteditor[1..-1]]
-          elsif cmd_exists? texteditor
-            return texteditor
-          end
+        texteditor = %w[/usr/bin/editor vi].detect do |texteditor|
+          cmd_exists?(texteditor)
+        end
+
+        if texteditor
+          return texteditor
+        else
+          raise "Could not determine you favorite CLI text editor. Use `sudo update-alternatives --config editor` to set your preference."
         end
       end
 


makmic:
A few years ago, geordi did actually look at `$EDITOR`, but there was a PR arguing against it:

https://github.com/makandra/geordi/pull/64

codener:
Ok, didn't know that one! However, I still find the "update-alternatives" command cumbersome.

What about whitelisting editors? Check `ENV['EDITOR']`. If it ends with vi, vim, or nano, (others?) we'll use that. Otherwise, fail with:
```
x Unsupported editor #{editor.inspect}.

Please choose one of #{editors.join('/')}, then run again with: EDITOR=vim geordi drop-databases 
```

makandra-hq:
ping

makmic:
Hm, why should we (the gem) keep a list of editors at all? For my part, I'd be fine if the user uses `hello-kitty-vim`. I still think the proposed patch is a good way forward

codener:
Well, I have never run `update-alternatives`, but /usr/bin/editor points to `nano`. I assume users will never run into the proposed error message.

What about printing a hint instead:
```
> Assuming your local user has permission to drop databases. Run with `--sudo` to use sudo.
> Select allowlist editor with: sudo update-alternatives --config editor
```
We should only print the hint if `/usr/bin/editor` exists. Otherwise, the user is locked to `vi` anyways.

wdyt?

makmic:
>  I assume users will never run into the proposed error message.

Uh, you are totally right! The proposal seems sensible to me

codener:
Fine :) Would you then update your PR, or should our team finish it?

makmic:
I'm unlikely to find time for this until July. I'd be thankful if your team could pick this up :)

PR #232, opened 2025-02-03
Fix parsing of server directives including bracket

chrkau:
Currently `server` directives including brackets will not be picked up by geordi. This should fix it.

### DIFF
--- a/lib/geordi/capistrano_config.rb
+++ b/lib/geordi/capistrano_config.rb
@@ -1,3 +1,5 @@
+require 'geordi/capistrano_config_parser'
+
 module Geordi
   class CapistranoConfig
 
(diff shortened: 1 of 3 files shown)


lib/geordi/capistrano_config.rb
```diff
@@ -11,11 +11,11 @@ def initialize(stage)
 
     def user(server)
       cap2user = deploy_info[/^\s*set\s*:user,\s*['"](.*?)['"]/, 1]
-      cap2user || deploy_info[/^\s*server\s*['"]#{server}['"],.*user.{1,4}['"](.*?)['"]/, 1]
+      cap2user || deploy_info[/^\s*server\s*\(?\s*['"]#{server}['"],.*user.{1,4}['"](.*?)['"]/, 1]
```
codener, line 14: Should match fine as
```suggestion
      cap2user || deploy_info[/^\s*server\(?\s*['"]#{server}['"],.*user.{1,4}['"](.*?)['"]/, 1]
```
Same below.

codener (changes requested):
Thanks for the suggestion, and for tests! I am willing to merge this, but with a little more restrictive syntax.

chrkau:
Your comments above made me think. "Parsing" Ruby with regexes seems kinda error prone. So I switched out the logic with a real parser. That should get rid of quite a few edge cases the regex approach misses.

codener:
Well ... the regex has been working fine for years. While adding a real parser may be "the true solution", I see several issues here:
- We have 100+ more LoC for doing the same thing as before.
- The `CapistranoConfigParser` is, like all parsers, hard to understand ("parse" ;-) as a human. This makes maintenance hard.
- We have a new dependency (ripper).
- Geordi maintains compatibility with several versions of Ruby, and Capistrano. A single parser implementation may fall short to support all, (potentially) introducing a host of new issues. Several years ago, we were parsing the configuration with Capistrano and asking it for specific values, but having to support several different versions rendered that approach unfeasible. That's why we're using simple regexes.

Your previous implementation was quite right. Would you please restore it?

chrkau:
I get it - but I'm having way too much fun with Ripper over here. :)
The code is up for grabs.
Thanks for you time.

PR #233, opened 2025-02-20
Add support for Ruby 3.4

tecden:


.github/workflows/test.yml
```diff
@@ -13,14 +13,14 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - ruby: 2.5.7
-            gemfile: Gemfile
```
tecden, line 17: Due to changes in Ruby 3.4 standard libraries, some gems had to be updated. These are no longer compatible with Ruby 2.5.7, so I have removed Ruby 2.5.7 from the test matrix. 

Alternatively, we can think about using the Gemika gems.

codener: That's fine. We have dropped full support for Ruby < 2.7 anyways. 

However, please document this in the Gemspec: https://guides.rubygems.org/specification-reference/#required_ruby_version

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Compatible changes
 
 - Add support for Ruby 3.3
+- Add support for Ruby 3.4
+- Drop support for Ruby 2.5.7
```
codener, line 12: This is a breaking change; please move to appropriate section.

Gemfile
```diff
@@ -10,4 +10,4 @@ gem 'highline'
 gem 'parallel_tests'
 gem 'launchy'
 gem 'pry-byebug'
-gem 'cucumber', '~> 4'
+gem 'cucumber', '~> 9'
```
codener, line 13: Maybe a version between 4 and 9 would have done as well ;) But since we're dropping support for 2.5 anyway, it's fine.

codener (approved):
Fine. Please apply the two changes, then you may release. Should get a new major version IMHO :thinking:

Gemfile
tecden: yeah, i thought, if we're dropping support for ruby 2.5, I might as well upgrade the gems to the highest possible version (as long as nothings has to be changed)

Question out of pure curiosity: Is there anything in favor of going with a lower version in this case?

PR #239, opened 2025-08-07
Enable irb flags for local geordi config and prioritize irb flags from config above default behavior

brunosedler:


lib/geordi/settings.rb
```diff
@@ -26,7 +26,8 @@ def initialize
 
     # Global settings
     def irb_flags
-      @global_settings['irb_flags']
+      return [@local_settings['irb_flags'].to_s, :local] if @local_settings.key? 'irb_flags'
+      [@global_settings['irb_flags'].to_s, :global] if @global_settings.key? 'irb_flags'
```
codener, line 30: Would be more idiomatic as:
```suggestion
      if @local_settings.key? 'irb_flags'
        [@local_settings['irb_flags'].to_s, :local] 
      elsif @global_settings.key? 'irb_flags'
        [@global_settings['irb_flags'].to_s, :global] 
      end
```

codener (approved):
Thanks for pushing this through!

PR #246, opened 2026-04-14
[W-389] 'geordi deploy': Identify actually deployed commits

LisaHader:


spec/util_spec.rb
```diff
@@ -58,6 +58,59 @@
 
   end
 
+  describe '.determine_issues_to_move' do
+    let(:source_branch) { 'feature/W-123-my-feature' }
+    let(:target_branch) { 'master' }
+    let(:target_stage) { 'staging' }
+
+    before do
+      allow(described_class).to receive(:sleep) # prevent wait time in errors paths for faster test execution
```
codener: 👌

lib/geordi/util.rb
```diff
@@ -220,6 +220,34 @@ def rspec_path?(path)
         %r{(^|\/)spec|_spec\.rb($|:)}.match?(path)
       end
 
+      def determine_issues_to_move(source_branch, target_branch, target_stage)
+        commit_messages = []
+        if CapistranoConfig.task_defined?('app:revision')
+          revision = CapistranoConfig.current_app_revisions(target_stage)
+          n = 3
```
codener: Please assign a meaningful name that is understandable without reading the surrounding code. E.g. `sleep_seconds`.

lib/geordi/util.rb
```diff
@@ -220,6 +220,34 @@ def rspec_path?(path)
...
+            puts "Continuing without moving Linear issues in #{n} seconds."
+            sleep(n)
+          else
+            # there is one consistent revision deployed across all target servers => Linear issues can be moved
+            commit_messages = Git.commits_after_revision(revision.first, source_branch)
+          end
+        else
+          Interaction.warn 'Missing Capistrano task "app:revision". See `geordi help deploy`.'
+          puts 'Without it, Geordi can only consider pushed commits for moving Linear issues.'
+          commit_messages = Git.commits_between(source_branch, target_branch)
```
codener: Actually, `commits_after_revision` is the same as `commits_between`.

L238 should do (simplified) "git diff revision..source_branch", L243 "git diff origin/target_branch..source_branch". (The Git command in `commits_after_revision` is not 100% correct anyways: it ignores possibly unpushed changes in `source_branch` locally, and there should not be a space after `..`.)

lib/geordi/util.rb
```diff
@@ -220,6 +220,34 @@ def rspec_path?(path)
...
 
+      def determine_issues_to_move(source_branch, target_branch, target_stage)
+        commit_messages = []
+        if CapistranoConfig.task_defined?('app:revision')
+          revision = CapistranoConfig.current_app_revisions(target_stage)
+          n = 3
+          if revision.empty?
+            Interaction.warn 'Could not read a server revision from app:revision output.'
+            puts "Continuing without moving Linear issues in #{n} seconds."
+            sleep(n)
```
codener: If there is no revision on the server, the deploy should fail. This must be fixed.

lib/geordi/util.rb
```diff
@@ -220,6 +220,34 @@ def rspec_path?(path)
...
+          elsif revision.size > 1
+            Interaction.warn 'Currently deployed revision differs on servers.'
+            puts "Continuing without moving Linear issues in #{n} seconds."
+            sleep(n)
```
codener: This should need confirmation (no sleep). It's a bad state that the user must be informed about (although she is fixing it right now =).

codener (changes requested):
Thank you! I have one simplification suggestion, and two workflow changes.

LisaHader:
@codener Thanks for the review! I implemented your suggested changes

lib/geordi/git.rb
```diff
@@ -41,18 +41,10 @@ def default_branch
         default_branch || 'master'
       end
 
-      def commits_between(source_branch, target_branch)
+      def commits_since(base, source_branch)
```
codener: What is the concept behind this naming: "since", "base" and "source_branch"? Isn't this method precisely listing commits between two refs? I'd prefer `commits_between` still, but let's discuss.

lib/geordi/util.rb
```diff
@@ -224,23 +224,20 @@ def determine_issues_to_move(source_branch, target_branch, target_stage)
...
           if revision.empty?
-            Interaction.warn 'Could not read a server revision from app:revision output.'
-            puts "Continuing without moving Linear issues in #{n} seconds."
-            sleep(n)
+            Interaction.fail 'Could not read a server revision from app:revision output.'
           elsif revision.size > 1
             Interaction.warn 'Currently deployed revision differs on servers.'
-            puts "Continuing without moving Linear issues in #{n} seconds."
-            sleep(n)
+            Interaction.confirm_or_cancel('Continue deployment without moving Linear issues?')
```
codener, line 232: 👌

codener (commented):
Thanks for the update! One more remark.

lib/geordi/git.rb
LisaHader: I initially felt that `since` was the better choice, probably because I was focused on the 'since the last revision' logic. However, after re-evaluating it with a few days of distance, I agree that between_commits is the more precise naming. I have updated the method accordingly. Thank you for pointing that out

codener (approved):
LGTM!

PR #247, opened 2026-06-15
[W-389] `geordi deploy`: only announce Linear issues that will actually be moved

LisaHader:


lib/geordi/commands/deploy.rb
```diff
@@ -127,9 +127,11 @@ def deploy(target_stage = nil)
 
   if move_issues
     linear_issue_ids, relevant_commits = Util.determine_issues_to_move(source_branch, target_branch, target_stage)
+    linear_issue_ids = linear_client.filter_existing_issue_ids(linear_issue_ids)
+    relevant_commits = LinearClient.filter_by_issue_ids(relevant_commits, linear_issue_ids)
```
codener: Could all of this live in `Util.determine_issues_to_move`? Seems a bit looped right now ♻️

LisaHader: I originally decided on that implementation because I wanted to avoid having to fetch all of the linear issues twice (since they are cached in an instance variable and we are reusing them again in `move_issues_to_state).` 

However, I agree that this is probably not be the best solution. I moved this into the `Util` method and pass the existing `linear_client` instance to it now to still be able to use the caching mechanism.

codener (approved):
Thanks for the update! I think the code is better now.

I do not totally like that the Linear client instance is passed to a utility method, but I could not come up with anything better. The impact is small. Let's keep this.

lib/geordi/linear_client.rb
```diff
@@ -83,12 +83,18 @@ def move_issues_to_state(issue_identifiers, state)
...
 
       issues = fetch_linear_issues # This only retrieves issues for the configured linear team ids
-      state_ids_by_team_id = state_ids_by_team_id(state)
 
-      issue_identifiers.each do |identifier|
-        issue = issues.find { |i| i['identifier'] == identifier }
+      # Only look up states for teams that actually have issues being moved
+      # This avoids warning about missing states in teams with no relevant issues.
+      relevant_issues = issue_identifiers.filter_map { |id| issues.find { |i| i['identifier'] == id } }
+      return if relevant_issues.empty?
```
codener, line 90: This is the new line, right? I am wondering if this is the right place to decide. In `move_issues_to_state`, shouldn't we already have reliably decided which issues to move? After all, didn't we announce _before the deploy_ what we plan to move? I think, that is the place where we actually decide. 

wdty?

LisaHader: Yes, this is the new line. Sorry, I already stashed some of the commits. We never call this method, if no relevant issues are present:

```
if move_issues && !linear_issue_ids.empty? # nothing gets moved when there are no relevant linear issue ids
    linear_client.move_issues_to_state(linear_issue_ids, target_state)
    settings.persist_linear_state_after_deploy(target_stage, target_state)
end
```

We already check if all the `linear_issue_ids` are present in `Util.determine_issues_to_move`, which then returned the linear_issue_ids we use for the check above. The issues there were cached, so no changes should be occurring in the meantime and the result is the same.

We should therefore never actually reach the case that no issues are left to move after the `filter_map.` I just added the line as an additional safety guard.

codener: Ok!

PR #248, opened 2026-06-16
do not add Linear URL to commit message

kratob:
Addint the Linear URL to commit messages is actively harmful, because it breaks Linear <-> Gitlab integration.

When a MR is created, Linear is supposed to post a linkback that includes the description of the Linear issue. This is very useful, especially for Gitlab code reviews. However, if the description of the MR *already contains the Linear URL*, Linear decides that the linkback is already present and will never post it.

### DIFF
--- a/features/commit.feature
+++ b/features/commit.feature
@@ -10,14 +10,6 @@ Feature: Creating a git commit from a Linear issue
     Then the output should contain "No issues to offer."
 
 
-  Scenario: It creates a commit with the linear issue url as description
-    Given I have staged changes
-
-    When I run `geordi commit` interactively
-      # No optional message
-    And I type ""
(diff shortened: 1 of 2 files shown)


codener:
Oh no. Had seemed so handy.

features/commit.feature
```diff
@@ -10,13 +10,13 @@ Feature: Creating a git commit from a Linear issue
     Then the output should contain "No issues to offer."
 
 
-  Scenario: It creates a commit with the linear issue url as description
+  Scenario: It creates a commit with the linear issue description as description
```
codener: I think this scenario must be deleted, as it is all about the code you deleted. It's title is actually wrong and misleading now.

codener (changes requested):
If you agree with my remark, you may merge after deleting the test.

